PR #6, opened 2022-08-23
Add script for generating manifests

reachjainrahul:
Changes include 
- Kustomization version upgrade to 4.4.1
- scripts for generating manifests files

### DIFF
--- a/hack/generate-manifest.sh
+++ b/hack/generate-manifest.sh
@@ -1,0 +1,87 @@
+#!/usr/bin/env bash
+
+# Copyright 2022 Antrea Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


hack/generate-manifest.sh
```diff
@@ -0,0 +1,87 @@
...
+set -eo pipefail
+
+function echoerr {
+    >&2 echo "$@"
+}
+
+_usage="Usage: $0 [--help|-h]
+Generate a YAML manifest for Nephe using Kustomize and print it to stdout.
+        --help    | -h                      Print this message and exit
+Environment variables IMG_NAME and IMG_TAG must be set when release mode is enabled.
```
Anandkumar26, line 26: nit: 
```suggestion
Environment variables IMG_NAME and IMG_TAG must be set.
```

Anandkumar26 (approved):
Minor comment. Can be addressed in a follow up patch.

hack/generate-manifest.sh
reachjainrahul: Ok.

PR #15, opened 2022-08-24
resolve minor issues in k8s secret integration with CPA

shenmo3:
refactor code and import and messages that are left when implementing cloud credentials to use k8s secret.

### DIFF
--- a/config/samples/cloudaccount_aws.yaml
+++ b/config/samples/cloudaccount_aws.yaml
@@ -1,6 +1,7 @@
 # Add AWS Account and Onboard VPC
 # To get base64 encoded json string for secret credential, run:
-# echo '{"accessKeyId": "YOUR_AWS_ACCESS_KEY_ID", "accessKeySecret": "YOUR_AWS_ACCESS_KEY_SECRET", "roleArn": "YOUR_AWS_IAM_ROLE_ARN", "externalId": "IAM_ROLE_EXTERNAL_ID"}' | openssl base64
+# echo '{"accessKeyId": "YOUR_AWS_ACCESS_KEY_ID", "accessKeySecret": "YOUR_AWS_ACCESS_KEY_SECRET", "roleArn": "YOUR_AWS_IAM_ROLE_ARN",
+# "externalId": "IAM_ROLE_EXTERNAL_ID"}' | openssl base64
 apiVersion: v1
 kind: Secret
 metadata:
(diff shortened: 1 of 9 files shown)


pkg/cloud-provider/cloudapi/azure/azure_account_mgmt.go
```diff
@@ -80,10 +79,6 @@ func compareAccountCredentials(accountName string, existing interface{}, new int
 
 // extractSecret extracts credentials from a Kubernetes secret.
 func extractSecret(c client.Client, s *v1alpha1.SecretReference) (*v1alpha1.AzureAccountCredential, error) {
-	if s == nil {
-		return nil, fmt.Errorf("secret reference not found")
```
bangqipropel, line 84: One thing here, we have a lot of error messages same in Azure and AWS here, like this `secret reference not found` , shall we add some suffix here to distinguish? maybe "secret reference not found in Azure" and secret reference not found in AWS" may be better?

config/samples/cloudaccount_aws.yaml
```diff
@@ -1,6 +1,7 @@
 # Add AWS Account and Onboard VPC
 # To get base64 encoded json string for secret credential, run:
-# echo '{"accessKeyId": "YOUR_AWS_ACCESS_KEY_ID", "accessKeySecret": "YOUR_AWS_ACCESS_KEY_SECRET", "roleArn": "YOUR_AWS_IAM_ROLE_ARN", "externalId": "IAM_ROLE_EXTERNAL_ID"}' | openssl base64
+# echo '{"accessKeyId": "YOUR_AWS_ACCESS_KEY_ID", "accessKeySecret": "YOUR_AWS_ACCESS_KEY_SECRET", "roleArn": "YOUR_AWS_IAM_ROLE_ARN",
+# "externalId": "IAM_ROLE_EXTERNAL_ID"}' | openssl base64
```
reachjainrahul, line 4: This might cause JSON parser error as discussed.

shenmo3: Done

pkg/cloud-provider/cloudapi/azure/azure_account_mgmt.go
shenmo3: As discussed offline, the errors will also print out stacktraces, so its not really hard to identify whether aws or azure is having problem

reachjainrahul (approved):

reachjainrahul (approved):

PR #19, opened 2022-08-26
unit test for Azure part

bangqipropel:
it is good for review, and coverage is 52%

### DIFF
--- a/pkg/cloud-provider/cloudapi/azure/azure_security_test.go
+++ b/pkg/cloud-provider/cloudapi/azure/azure_security_test.go
@@ -1,0 +1,515 @@
+// Copyright 2022 Antrea Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package azure
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"strings"
+
+	network "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2021-03-01/network"
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/seancfoley/ipaddress-go/ipaddr"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	"antrea.io/nephe/apis/crd/v1alpha1"
+	cloudcommon "antrea.io/nephe/pkg/cloud-provider/cloudapi/common"
+	"antrea.io/nephe/pkg/cloud-provider/securitygroup"
+)
+
+var _ = Describe("Azure Cloud Security", func() {
+	var (
+		testAccountNamespacedName       = &types.NamespacedName{Namespace: "namespace01", Name: "account01"}
+		testSubID                       = "SubID"
+		credentials                     = "credentials"
+		testClientID                    = "ClientID"
+		testClientKey                   = "ClientKey"
+		testTenantID                    = "TenantID"
+		testRegion                      = "eastus"
+		testRG                          = "testRG"
+		nsgID                           = "nephe-ag-nsgID"
+		atAsgID                         = "nephe-at-atapplicationsgID"
+		atAsgName                       = "atapplicationsgID"
+		agAsgID                         = "nephe-ag-agapplicationsgID"
+		testPriority              int32 = 100
+		testSourcePortRange             = "*"
+		testDestinationPortRange        = "*"
+		testPrivateIP                   = "0.0.0.0"
+		testProtocol                    = 6
+		testFromPort                    = 22
+		testToPort                      = 23
+		testCidrStr                     = "192.168.1.1/24"
+
+		testVnet01   = "testVnet01"
+		testVnetID01 = fmt.Sprintf("/subscriptions/%v/resourceGroups/%v/providers/Microsoft.Network/virtualNetworks/%v",
+			testSubID, testRG, testVnet01)
+		testVnet02   = "testVnet02"
+		testVnetID02 = fmt.Sprintf("/subscriptions/%v/resourceGroups/%v/providers/Microsoft.Network/virtualNetworks/%v",
+			testSubID, testRG, testVnet02)
+		testVnetID03 = fmt.Sprintf("/subscriptions/%v/resourceGroups/%v/providers/Microsoft.Network/virtualNetworks/%v",
+			testSubID, testRG, "testVnet03")
+
+		testVnetPeer01   = "testVnetPeer01"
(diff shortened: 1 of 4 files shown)


pkg/cloud-provider/cloudapi/azure/azure_security_test.go
```diff
@@ -0,0 +1,251 @@
...
+			mockCtrl                        *gomock.Controller
+			mockAzureServiceHelper          *MockazureServicesHelper
+			mockazureNwIntfWrapper          *MockazureNwIntfWrapper
+			mockazureNsgWrapper             *MockazureNsgWrapper
+			mockazureAsgWrapper             *MockazureAsgWrapper
+			mockazureVirtualNetworksWrapper *MockazureVirtualNetworksWrapper
+			mockazureResourceGraph          *MockazureResourceGraphWrapper
+			mockazureService                *MockazureServiceClientCreateInterface
+		)
+		_ = account
```
shenmo3, line 75: This shouldn't be needed, can you remove it? Also in azure_test.go there is a similar instance at line 76, can you remove that too? These should be the left over thing we forgot to remove in the K8s secret PR.

pkg/cloud-provider/cloudapi/azure/azure_security_test.go
```diff
@@ -0,0 +1,404 @@
...
+		testClientID                    = "ClientID"
+		testClientKey                   = "ClientKey"
+		testTenantID                    = "TenantID"
+		testRegion                      = "eastus"
+		testRG                          = "testRG"
+		nsgID                           = "nephe-ag-nsgID"
+		atAsgID                         = "nephe-at-atapplicationsgID"
+		atAsgName                       = "atapplicationsgID"
+		agAsgID                         = "nephe-ag-agapplicationsgID"
+		testPriority              int32 = 100
```
reachjainrahul, line 53: do you need to set this?

pkg/cloud-provider/cloudapi/azure/azure_security_test.go
```diff
@@ -0,0 +1,404 @@
...
+		testCidrStr                     = "192.168.1.1/24"
+
+		testVnet01   = "testVnet01"
+		testVnetID01 = fmt.Sprintf("/subscriptions/%v/resourceGroups/%v/providers/Microsoft.Network/virtualNetworks/%v",
+			testSubID, testRG, testVnet01)
+		testVnet02   = "testVnet02"
+		testVnetID02 = fmt.Sprintf("/subscriptions/%v/resourceGroups/%v/providers/Microsoft.Network/virtualNetworks/%v",
+			testSubID, testRG, testVnet02)
+
+		testVnetPeer01   = "testVnetPeer01"
```
reachjainrahul, line 69: lets avoid Peering test for now

pkg/cloud-provider/cloudapi/azure/azure_security_test.go
```diff
@@ -0,0 +1,404 @@
...
+			err = c.AddProviderAccount(fakeClient, account)
+			Expect(err).Should(BeNil())
+			selector.Spec.VMSelector = vmSelector
+			err = c.AddAccountResourceSelector(testAccountNamespacedName, selector)
+			Expect(err).Should(BeNil())
+
+			accCfg, _ := c.cloudCommon.GetCloudAccountByName(testAccountNamespacedName)
+			serviceConfig, _ := accCfg.GetServiceConfigByName(azureComputeServiceNameCompute)
+
+			vnetIDs := make(map[string]struct{})
```
reachjainrahul, line 229: generally its not a good idea to combine everything in before test, unless its common for all. can we break this into smaller funcs/modules and explicitly call from the test. For example, define a function like setupAccount, addSelector etc.

pkg/cloud-provider/cloudapi/azure/azure_security_test.go
```diff
@@ -0,0 +1,404 @@
...
+							webAddressGroupIdentifier03,
+						},
+					},
+				}
+
+				err := c.UpdateSecurityGroupRules(webAddressGroupIdentifier03, ingressrules, egressrules)
+				Expect(err).Should(BeNil())
+			})
+
+			It("Should update Security rules for Peerings", func() {
```
reachjainrahul, line 344: For now, remove peering test. We will add it when some coverage on integration tests too

pkg/cloud-provider/cloudapi/azure/azure_security_test.go
```diff
@@ -0,0 +1,404 @@
...
+						Name: *webAddressGroupIdentifier02,
+					},
+				}
+
+				err := c.UpdateSecurityGroupMembers(webAddressGroupIdentifier01, members, false)
+				Expect(err).Should(BeNil())
+			})
+		})
+
+		Context("UpdateSecurityRules", func() {
```
reachjainrahul, line 304: could you add failure tests too

pkg/cloud-provider/cloudapi/azure/azure_test.go
```diff
@@ -244,7 +261,160 @@ var _ = Describe("Azure", func() {
...
+			err = c.AddAccountResourceSelector(testAccountNamespacedName, selector)
+			Expect(err).Should(BeNil())
+
+			accCfg, _ := c.cloudCommon.GetCloudAccountByName(testAccountNamespacedName)
+			serviceConfig, _ := accCfg.GetServiceConfigByName(azureComputeServiceNameCompute)
+			filters := serviceConfig.(*computeServiceConfig).computeFilters[selector.Name]
+			Expect(filters).To(Equal(expectedQueryStrs))
+		})
+
+		It("Should match expected filter - single VM Name only match", func() {
```
reachjainrahul, line 349: whats the difference between this test and test at line 265?

pkg/cloud-provider/cloudapi/azure/azure_test.go
```diff
@@ -244,7 +261,160 @@ var _ = Describe("Azure", func() {
 			filters := serviceConfig.(*computeServiceConfig).computeFilters[selector.Name]
 			Expect(filters).To(Equal(expectedQueryStrs))
 		})
+
+		It("Should match expected filter - single VMID only match", func() {
```
reachjainrahul, line 265: Try to add all combination. 
1) VNETID match
2) VMID match
3) VM Name Match
4) VNETID + VMID
5) VNETID + VMName
6) Update and rollback. Like start with VNETID, then add VM Match and then remove VM Match... and so on

pkg/cloud-provider/cloudapi/azure/azure_security_test.go
bangqipropel: I think nsgrule need this as int32

pkg/cloud-provider/cloudapi/azure/azure_security_test.go
bangqipropel: deleted

pkg/cloud-provider/cloudapi/azure/azure_security_test.go
bangqipropel: deleted

pkg/cloud-provider/cloudapi/azure/azure_security_test.go
bangqipropel: added

pkg/cloud-provider/cloudapi/azure/azure_security_test.go
bangqipropel: I think for asg and nsg part they can use for multiple times, it should be in the Return() part, and other create and addaccount part, it is always return an error in one line and then need a check error, I think this part also mostly the same steps in the aws tests

pkg/cloud-provider/cloudapi/azure/azure_test.go
bangqipropel: Here is the problem now, seems like our RemoveSelector is not completed, so we cannot do the rollback now

bangqipropel:
After removing the Peer part, and added some other tests, now coverage is 50.7%

bangqipropel:
addressed the comments and as discussed, added the peering part back, new coverage is 56.1%.

bangqipropel:
@reachjainrahul Updated last week and sent the email, please review it, thanks

reachjainrahul (approved):
/LGTM

PR #21, opened 2022-08-29
fix docs wording and add tag deployment

shenmo3:
## Changes
1. Fix word cases and typos in docs.
2. Add instruction for Nephe deployment using tagged yamls.


docs/getting-started.md
```diff
@@ -58,17 +66,18 @@ to [the AKS installation guide](aks-installation.md).
...
 ## Importing Cloud VMs
 
-To manage security policies of Public Cloud VMs, we need to first import
-the target VMs onto `Nephe Controller`. The following section explains
-how to set up access to public cloud account, select target VMs, and
-import VMs into the K8s cluster as VirtualMachine CRs.
+To manage security policies of Public Cloud VMs, we need to first import the
+target VMs onto the `Nephe Controller`. The following section explains how to
+set up access to public cloud account, select target VMs, and import VMs into
+the K8s cluster as `VirtualMachine` CRs.
```
jianjuns, line 72: Since not just CloudProviderAccount but also CloudEntitySelector need to be created in the Namespace, probably we should introduce the Namespace here, instead of in the CloudEntitySelector section. How about "and import VMs into a Namespace as`VirtualMachine` CRs. The sample commands import VMs into the `sample-ns` Namespace."?

docs/getting-started.md
```diff
@@ -58,17 +66,18 @@ to [the AKS installation guide](aks-installation.md).
...
 
 ### CloudProviderAccount
 
 To import cloud VMs, user needs to configure a `CloudProviderAccount` CR, with
```
jianjuns, line 76: "Before importing cloud VMs, you need to.."?

docs/getting-started.md
shenmo3: Done with slight modification.

docs/getting-started.md
shenmo3: Done with slight modification. I choose to use `we` as I think that sounds more friendly.

jianjuns (approved):

reachjainrahul (approved):

PR #26, opened 2022-09-16
fix aws dependency violation error

shenmo3:
## Description

This PR fixes the dependency violation error for AWS. When ANP rule and addressGroup changes, cloud controller attempts to delete previous addressGroup before checking if the associated security rule in appliedToGroup is deleted. This result in a dependency violation error returned by AWS. The deletion eventually succeed on retry after the security rule is removed. This PR adds such security rule association and checks before deleting addressGroup, putting it into a pending state. No user behavior is changed

## Changes

1. Add a map in appliedToSG to track its rule referenced addressGroups. Add an index for the reference.
2. Block addressGroup delete if any reference exist.
3. Fix some import and comment format.

### DIFF
--- a/pkg/controllers/cloud/networkpolicy.go
+++ b/pkg/controllers/cloud/networkpolicy.go
@@ -535,12 +535,13 @@ func (s *securityGroupImpl) deleteImpl(c cloudSecurityGroup, membershipOnly bool
 		r.pendingDeleteGroups.Add(guName, &pendingGroup{refCnt: new(int)})
 	}
 	if s.state != securityGroupStateGarbageCollectState {
-		sgs, err := r.networkPolicyIndexer.ByIndex(indexKey, s.id.Name)
+		nps, err := r.networkPolicyIndexer.ByIndex(indexKey, s.id.Name)
 		if err != nil {
 			return fmt.Errorf("get networkpolicy indexer with index=%v, name=%v: %w", indexKey, s.id.Name, err)
 		}
-		if len(sgs) != 0 {
-			r.Log.V(1).Info("Deleting SecurityGroup pending", "Name", s.id, "MembershipOnly", membershipOnly)
+		if len(nps) != 0 {
+			r.Log.V(1).Info("Deleting SecurityGroup pending, in use by networkpolicies", "Name", s.id,
+				"MembershipOnly", membershipOnly, "anpNum", len(nps))
 			s.deletePending = true
 			return nil
 		}
@@ -550,9 +551,8 @@ func (s *securityGroupImpl) deleteImpl(c cloudSecurityGroup, membershipOnly bool
 				return fmt.Errorf("get appliedTo indexer with name=%v: %w", s.id.String(), err)
 			}
 			if len(refs) != 0 {
-				r.Log.V(1).Info(
-					"Deleting SecurityGroup pending", "Name", s.id, "MembershipOnly", membershipOnly, "refs", len(refs),
-				)
+				r.Log.V(1).Info("Deleting SecurityGroup pending, referenced by appliedTo groups", "Name", s.id,
+					"MembershipOnly", membershipOnly, "refNum", len(refs))
 				s.deletePending = true
 				return nil
 			}
@@ -801,9 +801,9 @@ func (a *addrSecurityGroup) notifyNetworkPolicyChange(r *NetworkPolicyReconciler
 // appliedToSecurityGroup contains information to create a cloud appliedToSecurityGroup.
 type appliedToSecurityGroup struct {
 	securityGroupImpl
-	hasRules   bool
-	hasMembers bool
-	references map[string]bool
+	hasRules     bool
+	hasMembers   bool
+	addrGroupRef map[string]bool
 }
 
 // newAddrAppliedGroup creates a new addSecurityGroup from Antrea AddressGroup membership.
@@ -907,7 +907,7 @@ func (a *appliedToSecurityGroup) computeRules(nps []interface{}) ([]*securitygro
 	return deduplicateIngressRules(irules), deduplicateEgressRules(erules)
 }
 
-// updateAddrGroupReference updates appliedTo group references and notifies removed addrGroups that rules referencing them is removed.
+// updateAddrGroupReference updates appliedTo group addrGroupRef and notifies removed addrGroups that rules referencing them is removed.
 func (a *appliedToSecurityGroup) updateAddrGroupReference(r *NetworkPolicyReconciler) error {
 	// get latest irules and erules
 	nps, err := r.networkPolicyIndexer.ByIndex(networkPolicyIndexerByAppliedToGrp, a.id.Name)
@@ -916,7 +916,7 @@ func (a *appliedToSecurityGroup) updateAddrGroupReference(r *NetworkPolicyReconc
 	}
 	irules, erules := a.computeRules(nps)
 
-	// combine irules and erules to get latest addrSG references.
+	// combine irules and erules to get latest addrGroupRef.
 	references := make(map[string]bool)
 	for _, rule := range irules {
 		for _, sg := range rule.FromSecurityGroups {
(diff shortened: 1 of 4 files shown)


pkg/controllers/cloud/networkpolicy.go
```diff
@@ -544,6 +544,19 @@ func (s *securityGroupImpl) deleteImpl(c cloudSecurityGroup, membershipOnly bool
...
 		}
+		if membershipOnly {
+			refs, err := r.appliedToSGIndexer.ByIndex(appliedToIndexerByAddrGroupRef, s.id.String())
+			if err != nil {
+				return fmt.Errorf("get appliedTo indexer with name=%v: %w", s.id.String(), err)
+			}
+			if len(refs) != 0 {
+				r.Log.V(1).Info(
+					"Deleting SecurityGroup pending", "Name", s.id, "MembershipOnly", membershipOnly, "refs", len(refs),
+				)
```
reachjainrahul, line 555: nit: end this in previous line

pkg/controllers/cloud/networkpolicy.go
```diff
@@ -772,13 +785,13 @@ func (a *addrSecurityGroup) notifyNetworkPolicyChange(r *NetworkPolicyReconciler
 	if !a.deletePending {
 		return
 	}
-	sgs, err := r.networkPolicyIndexer.ByIndex(networkPolicyIndexerByAddrGrp, a.id.Name)
+	nps, err := r.networkPolicyIndexer.ByIndex(networkPolicyIndexerByAddrGrp, a.id.Name)
 	if err != nil {
 		r.Log.Error(err, "Get networkPolicy indexer", a.id.Name, err, "indexKey", networkPolicyIndexerByAddrGrp)
```
reachjainrahul, line 790: I know its legacy code, but is there any reason to proceed further if err occurs?

pkg/controllers/cloud/networkpolicy.go
```diff
@@ -790,6 +803,7 @@ type appliedToSecurityGroup struct {
 	securityGroupImpl
 	hasRules   bool
 	hasMembers bool
+	references map[string]bool
```
reachjainrahul, line 806: references is generic. Can we change to addressGroupRef ??

pkg/controllers/cloud/networkpolicy.go
```diff
@@ -949,6 +1030,9 @@ func (a *appliedToSecurityGroup) notify(op securityGroupOperation, status error,
 		a.hasMembers = true
 	case securityGroupOperationUpdateRules:
 		// AppliedToSecurityGroup added rules, now add members.
```
reachjainrahul, line 1032: Move the comment

pkg/controllers/cloud/networkpolicy.go
```diff
@@ -864,24 +878,91 @@ func (a *appliedToSecurityGroup) updateRules(r *NetworkPolicyReconciler) error {
...
+		}
+	}
+	for _, rule := range erules {
+		for _, sg := range rule.ToSecurityGroups {
+			references[sg.String()] = true
+		}
+	}
+	// compute references removed from previous.
+	removedRef := make([]string, 0)
+	for oldRef := range a.references {
```
reachjainrahul, line 933: Where are you updating a.references???

pkg/controllers/cloud/networkpolicy.go
```diff
@@ -959,6 +1043,12 @@ func (a *appliedToSecurityGroup) notify(op securityGroupOperation, status error,
 		if a.hasRules {
 			return a.updateRules(r)
 		}
+	case securityGroupOperationDelete:
```
reachjainrahul, line 1046: Why is this required??

pkg/controllers/cloud/networkpolicy.go
```diff
@@ -864,24 +878,91 @@ func (a *appliedToSecurityGroup) updateRules(r *NetworkPolicyReconciler) error {
...
-	ch := securitygroup.CloudSecurityGroup.UpdateSecurityGroupRules(&a.id, irules, erules)
-	go func() {
-		err := <-ch
-		r.cloudResponse <- &securityGroupStatus{sg: a, op: securityGroupOperationUpdateRules, err: err}
-	}()
+	return deduplicateIngressRules(irules), deduplicateEgressRules(erules)
+}
+
+// updateAddrGroupReference updates appliedTo group references and notifies removed addrGroups that rules referencing them is removed.
+func (a *appliedToSecurityGroup) updateAddrGroupReference(r *NetworkPolicyReconciler) error {
```
reachjainrahul, line 911: Can you verify what happens when addressGroup and AppliedToGroup are updated together. Would it still preserve your references. You can try it out with just updating appliedToGroup and verify if appliedTo update keeps the references intact

shenmo3: Checked it works. It doesn't preserve the reference, but it will recompute the reference when new AT is created and update rule called.

pkg/controllers/cloud/networkpolicy.go
shenmo3: If AT changed but np doesn't, either with some corner case or event received out of order from antrea (delete NP then delete AT), the controller never notify addrGroups since AT delete doesn't call updateRules.

pkg/controllers/cloud/networkpolicy.go
shenmo3: I prefer have comments at the top of statements, clearly indicate the functionality of the following statements. I have updated the comment to reflect the latest changes

pkg/controllers/cloud/networkpolicy.go
shenmo3: Done

pkg/controllers/cloud/networkpolicy.go
shenmo3: No particular reason. We can try returning on error and see if it breaks anything. My guess is indexer error rarely happens and is probably never tested.

pkg/controllers/cloud/networkpolicy.go
shenmo3: Done

pkg/controllers/cloud/networkpolicy.go
```diff
@@ -788,8 +801,9 @@ func (a *addrSecurityGroup) notifyNetworkPolicyChange(r *NetworkPolicyReconciler
 // appliedToSecurityGroup contains information to create a cloud appliedToSecurityGroup.
 type appliedToSecurityGroup struct {
 	securityGroupImpl
-	hasRules   bool
-	hasMembers bool
+	hasRules     bool
+	hasMembers   bool
+	addrGroupRef map[string]bool
```
reachjainrahul, line 806: nit: addrGroupRefs

pkg/controllers/cloud/networkpolicy.go
```diff
@@ -864,24 +878,91 @@ func (a *appliedToSecurityGroup) updateRules(r *NetworkPolicyReconciler) error {
...
+			references[sg.String()] = true
+		}
+	}
+	for _, rule := range erules {
+		for _, sg := range rule.ToSecurityGroups {
+			references[sg.String()] = true
+		}
+	}
+	// compute addrGroupRef removed from previous.
+	removedRef := make([]string, 0)
```
reachjainrahul, line 932: removedRefs? and update comment accordingly

pkg/controllers/cloud/networkpolicy.go
```diff
@@ -948,17 +1029,27 @@ func (a *appliedToSecurityGroup) notify(op securityGroupOperation, status error,
 	case securityGroupOperationUpdateMembers:
 		a.hasMembers = true
 	case securityGroupOperationUpdateRules:
-		// AppliedToSecurityGroup added rules, now add members.
+		// AppliedToSecurityGroup added rules, now update addrGroup references and add members.
```
reachjainrahul, line 1032: Generally, at level comments like this are used for function to share the intent. this is fine for now. but in future if this case grows, this kind of top level comments will go thru iterations

pkg/controllers/cloud/networkpolicy.go
shenmo3: Done

pkg/controllers/cloud/networkpolicy.go
shenmo3: Done

reachjainrahul (approved):

PR #29, opened 2022-09-20
Update rule realization for ANP

reachjainrahul:
Nephe controller combines all the ANP rules and apply them to cloud together. As such, all ANPs having same appliedToGroup will share the same fate. Here, with every ANP change, we walk thru all the ANPs sharing the same appliedTo and report combined rule realization statuses.

Note: As the ANPs are combined together, antrea controller will receive dup realization status for already realized rule if there are more than 1 ANP sharing the same appliedToGroup.

### DIFF
--- a/config/nephe.yml
+++ b/config/nephe.yml
@@ -454,6 +454,13 @@ rules:
   - list
   - watch
 - apiGroups:
+  - controlplane.antrea.io
+  resources:
+  - networkpolicies/status
+  verbs:
+  - create
+  - delete
+- apiGroups:
   - crd.antrea.io
   resources:
   - externalentities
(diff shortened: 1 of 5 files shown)


Anandkumar26 (approved):
/LGTM

config/nephe.yml
```diff
@@ -453,6 +453,13 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - controlplane.antrea.io
+  resources:
+  - networkpolicies/status
+  verbs:
+  - create
+  - delete
```
Anandkumar26, line 462: Do you need delete permission?

reachjainrahul: not with this change, but when antrea supports delete on ANP realization, no change would be required on RBAC

PR #31, opened 2022-09-22
integrate gc jobs into one

liu4480:
previously, we split gc into two jobs, it is better to integrate into one

Signed-off-by: Bin Liu <biliu@vmware.com>

### DIFF
--- a/ci/jenkins/jobs/projects.yaml
+++ b/ci/jenkins/jobs/projects.yaml
@@ -377,12 +377,19 @@
                       default-excludes: true
                       fingerprint: false
                       only-if-success: false
-      - '{test_name}-testbed-gc':
+      - '{test_name}-testbed-gc-on-all-nodes':
           test_name: nephe
-          node: 'nephe'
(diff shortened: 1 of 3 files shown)


ci/jenkins/jobs/projects.yaml
```diff
@@ -399,18 +406,5 @@
...
-          axes:
-            - axis:
-                type: slave
-                name: nodes
-                values:
-                  - nephe-test-1
-                  - nephe-test-2
-                  - nephe-test-3
           builders:
             - nephe-testbed-gc-on-all-nodes
```
edwardbadboy, line 410: This builder should be nephe-gc-cronjob, and we can delete builder nephe-testbed-gc-on-all-nodes from macros.yaml.

edwardbadboy (changes requested):

edwardbadboy (approved):

liu4480:
@reachjainrahul @Anandkumar26 can you help to review when you are available? thanks

reachjainrahul (approved):

PR #34, opened 2022-09-26
perform delta update on aws security rule

shenmo3:
## Description

This PR changes aws security rule update behavior from revoke all then apply all to a delta update, only deleting / adding necessary rules.

## Changes
1. Implement custom comparison function for IpPermissions. Many fields in IpPermission.UserIdGroupPair are not compared due to not used or aws added metadata.
2. Implement rule delta compute function between new rules and current rules.
3. Change ingress and egree realization to use add and delete list computed by delta function in authorize and revoke operation.
4. Added unit test for aws update security rules, coverage for aws 38.5% -> 50.8%.

### DIFF
--- a/pkg/cloud-provider/cloudapi/aws/aws_security.go
+++ b/pkg/cloud-provider/cloudapi/aws/aws_security.go
@@ -16,6 +16,7 @@ package aws
 
 import (
 	"fmt"
+	"reflect"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/cloud-provider/cloudapi/aws/aws_security.go
```diff
@@ -590,6 +593,82 @@ func (ec2Cfg *ec2ServiceConfig) getNepheControllerManagedSecurityGroupsCloudView
...
+	if protocol, ok := securitygroup.ProtocolNameNumMap[*p2Protocol]; ok {
+		p2Protocol = convertToIPPermissionProtocol(&protocol)
+	}
+	if *p1Protocol != *p2Protocol {
+		return false
+	}
+
+	groupIds := make(map[string]bool)
+	for _, id := range p1.UserIdGroupPairs {
+		groupIds[*id.GroupId] = true
```
reachjainrahul, line 661: I believe this is addressgroup AG ID. Can there be case we have more than 1 UserIdGroup Pair?

reachjainrahul (approved):

pkg/cloud-provider/cloudapi/aws/aws_security_test.go
```diff
@@ -192,25 +194,210 @@ var _ = Describe("AWS Cloud Security", func() {
 				Name: "Web",
 				Vpc:  testVpcID01,
 			}
-			input1 := testAwsBuildDescribeSecurityGroupInput(webAddressGroupIdentifier.Vpc,
+			input1 := constructDescribeSecurityGroupInput(webAddressGroupIdentifier.Vpc,
```
reachjainrahul, line 197: keep it EC2 only to be consistent with aws sdk

reachjainrahul (approved):

PR #40, opened 2022-10-11
Add integration test for antrea group

reachjainrahul:
Added integration test for ANP using appliedTo as
Antrea Groups rather than external entity selector. Test will first create Antrea Group and use that group as AppliedTo field in ANP


Anandkumar26 (approved):
/LGTM

PR #46, opened 2022-10-20
Normalizes ANP Object for Azure Cloud

reachjainrahul:
Azure Cloud doesn't handle / in any cloud resource name. So we normalize the object before calling cloud plugin. This is temporary workaround till https://github.com/antrea-io/antrea/issues/4264 issue is fixed in Antrea.

Signed-off-by: Rahul Jain <reachjainrahul@gmail.com>


Anandkumar26 (approved):
/LGTM

PR #51, opened 2022-10-25
Update rule realization for failure case too

reachjainrahul:
With antrea 1.9, networkpolicystatus support realization status and error msg field. Added the backend support in Nephe too to report realization for both success and failure case.

Signed-off-by: Rahul Jain <reachjainrahul@gmail.com>


Anandkumar26 (approved):
/LGTM

PR #61, opened 2022-10-31
fix gc job

liu4480:
gc job failed to run when it checks there is stale testbed, this is because it always checks if the testbed is in directory `terraform.tfstate.d`, this works when we deploy testbed and destroy testbed upon job finishes. But gc job does not know its original directory, so we store it in /var/lib/jenkins/, the terraform directory should be specified when calling destroy.sh.
Signed-off-by: Bin Liu <biliu@vmware.com>

### DIFF
--- a/ci/jenkins/nephe-gc.sh
+++ b/ci/jenkins/nephe-gc.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
-
 # Copyright 2022 Antrea Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


liu4480:
> @liu4480 could you update the PR with what was the issue and what is the fix?

thanks for the reminding, will update

ci/jenkins/nephe-gc.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-
+set -x
```
reachjainrahul, line 2: can you move this after copyright

ci/jenkins/nephe-gc.sh
```diff
@@ -20,7 +20,8 @@
 
 _usage="Usage: $0 [--goVcPassword <Password for VC>]
   --goVcPassword          Password to the user name for VC.
-  --terraform-dir         directory for terraform."
```
reachjainrahul, line 23: Capital letter, Directory

liu4480: done

ci/jenkins/nephe-gc.sh
liu4480: done

PR #64, opened 2022-11-03
Improve error message on ci setup fails

shenmo3:
## Description

For CI tests, if the setup or parameters errors out, the actual error message output can be confusing. This is due to `AfterSuite` running when some steps in `BeforeSuite` fails, either due to parameter error or setup issue. Because setup is not finished, `AfterSuite` clean up also encounters and throws error, causing confusion. This PR fixes two cases.

## Changes
1. Add checks for Nephe controller deployment and skip some checks if its not deployed.
2. Add checks for kubectl configuration and skip clean up if the cluster is not configured.
3. Fixed a formatting issue for namespace clean up.

### DIFF
--- a/test/integration/integration_suite_test.go
+++ b/test/integration/integration_suite_test.go
@@ -212,10 +212,17 @@ var _ = AfterSuite(func(done Done) {
 		close(done)
 		return
 	}
+
+	if kubeCtl == nil {
+		logf.Log.Info("No cluster found")
+		close(done)
+		return
+	}
+
 	var controllersCored *string
 	var err error
+	kubeCtl.SetContext(clusterContext)
 	cluster := clusterContext
-	kubeCtl.SetContext(cluster)
 	if len(cluster) == 0 {
 		cluster = "default"
 	}
@@ -223,8 +230,12 @@ var _ = AfterSuite(func(done Done) {
 	By(cluster + ": Check for controllers' restarts")
 	err = utils.CheckRestart(kubeCtl)
 	if err != nil {
-		logf.Log.Error(err, "error restarting nephe controller")
-		controllersCored = &cluster
+		if apierrors.IsNotFound(err) {
+			logf.Log.Info("nephe controller not deployed")
+		} else {
+			logf.Log.Error(err, "error restarting nephe controller")
+			controllersCored = &cluster
+		}
 	}
 
 	if controllersCored != nil {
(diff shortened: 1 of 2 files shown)


test/integration/integration_suite_test.go
```diff
@@ -212,19 +212,30 @@ var _ = AfterSuite(func(done Done) {
 		close(done)
 		return
 	}
+
+	if kubeCtl == nil {
+		logf.Log.Info("No cluster found")
```
reachjainrahul, line 217: why is this info message. isnt this an error.

test/integration/integration_suite_test.go
```diff
@@ -212,19 +212,30 @@ var _ = AfterSuite(func(done Done) {
...
 
 	By(cluster + ": Check for controllers' restarts")
 	err = utils.CheckRestart(kubeCtl)
 	if err != nil {
-		logf.Log.Error(err, "error restarting nephe controller")
-		controllersCored = &cluster
+		if apierrors.IsNotFound(err) {
+			logf.Log.Info("nephe controller not deployed")
+		} else {
+			logf.Log.Error(err, "error restarting nephe controller")
```
reachjainrahul, line 236: if there is an error, should we return?

reachjainrahul (commented):
Can we avoid AfterSuite, if BeforeSuite fails??

test/integration/integration_suite_test.go
shenmo3: Not really, this error indicates that the nephe controller has restarted during the test (pod restart count > 0). In this case we continue the AfterSuite clean up and collect support bundle or preserve setup later. At the end the test will error out.

test/integration/integration_suite_test.go
shenmo3: No this means kubectl is not configured, means setup is not complete, some error happened during BeforeSuite.

shenmo3:
> Can we avoid AfterSuite, if BeforeSuite fails??

I think we should run AfterSuite even if BeforeSuite fails, because some setup might already be there and some are not due to partially ran BeforeSuite. AfterSuite should try its best to clean up so the cluster won't be in a partially configured confusing state.

reachjainrahul (dismissed):
/LGTM

reachjainrahul (approved):
/LGTM

PR #83, opened 2023-01-04
Specify version for Azure/vnet/azurerm module

Anandkumar26:
Fixes the errror, argument "use_for_each" is required, but no definition was found


reachjainrahul (approved):
/LGTM

PR #90, opened 2023-01-12
Sync with cloud feature hardening - 1

archanapholla:
Issue1:
    Sync content member compare(includes NetworkInterface) always fails.
Fix:
    Name field in CloudResourceID holds the VM ID(resource ID), hence
    use the indexer meant for matching VM ID fix in VM CRs.
    While creating network interface CloudResource object, use
    the right id in CloudResourceID.

Issue2:
    Synchronization content is never received for AT SG.
Fix:
    An example of NSG in Azure is nephe-at-per-vnet-default-archana-test-vnet-1.
    Fix is related to string comaprision of SG name with "per-vnet-default".
    As this is not a direct comparison, use "Contains" instead of "Compare".

Issue3:
    Sync content is not received for AG SG when AT SG with same name exists.
Fix:
    Irrespective of AT group names, fix creates sync content for AG.
    Name collision between AG and AT group can arise when two ANP's
    vm1 AT -> vm2 AG and vm2 AT -> vm1 AG exist.

Issue4:
    Stale sync content received for AG SG which belong to unmanaged VPCs.
Fix:
    While building CloudResource Map for ASG, it is built only for managed VPC,
    however after cloud API call for getting all ASGs is made,
    while processing the result, all nephe created ASGs are taken into
    account even if they do not have any CloudResource member/vpc association.
    Add a check to make sure ASG is associated with managed VPC and
    skip ASGs associated with unmanaged VPCs.

### DIFF
--- a/pkg/cloud-provider/cloudapi/azure/azure_security.go
+++ b/pkg/cloud-provider/cloudapi/azure/azure_security.go
@@ -506,6 +506,7 @@ func getAsgsToAdd(asgs *[]network.ApplicationSecurityGroup, addrGroupNepheContro
 	return &asgsToKeep, updated
 }
 
+// processAndBuildATSgView creates synchronization content for AppliedTo SG.
 func (computeCfg *computeServiceConfig) processAndBuildAGSgView(
 	networkInterfaces []*networkInterfaceTable) ([]securitygroup.SynchronizationContent, error) {
 	nepheControllerAGSgNameToMemberCloudResourcesMap := make(map[string][]securitygroup.CloudResource)
@@ -549,6 +550,7 @@ func (computeCfg *computeServiceConfig) processAndBuildAGSgView(
 	return addressGroupSgEnforcedView, err
 }
 
+// processAndBuildAGSgView creates synchronization content for AdressGroup SG.
 func (computeCfg *computeServiceConfig) processAndBuildATSgView(networkInterfaces []*networkInterfaceTable) (
 	[]securitygroup.SynchronizationContent, error) {
 	nepheControllerATSgNameToMemberCloudResourcesMap := make(map[string][]securitygroup.CloudResource)
@@ -609,6 +611,7 @@ func (computeCfg *computeServiceConfig) processAndBuildATSgView(networkInterface
 	return appliedToSgEnforcedView, err
 }
 
+// getATGroupView creates synchronization content for NSGs created by nephe under managed VNETs.
 func (computeCfg *computeServiceConfig) getATGroupView(nepheControllerATSGNameToCloudResourcesMap map[string][]securitygroup.CloudResource,
 	perVnetNsgIDToNepheControllerATSGNameSet map[string]map[string]struct{}, nsgIDToVnetID map[string]string) (
 	[]securitygroup.SynchronizationContent, error) {
@@ -652,6 +655,7 @@ func (computeCfg *computeServiceConfig) getATGroupView(nepheControllerATSGNameTo
 	return enforcedSecurityCloudView, nil
 }
 
+// getAGGroupView creates synchronization content for ASGs created by nephe under managed VNETs.
 func (computeCfg *computeServiceConfig) getAGGroupView(nepheControllerAGSgNameToCloudResourcesMap map[string][]securitygroup.CloudResource,
 	asgIDToVnetID map[string]string) ([]securitygroup.SynchronizationContent, error) {
 	appSecurityGroups, err := computeCfg.asgAPIClient.listAllComplete(context.Background())
@@ -662,7 +666,7 @@ func (computeCfg *computeServiceConfig) getAGGroupView(nepheControllerAGSgNameTo
 	var enforcedSecurityCloudView []securitygroup.SynchronizationContent
 	for _, appSecurityGroup := range appSecurityGroups {
 		asgIDLowercase := strings.ToLower(*appSecurityGroup.ID)
-		// if Asg is not associated with a managed VPC, do not create sync content for it.
+		// if ASG is not associated with a managed VNET, do not create sync content for it.
 		if _, found := asgIDToVnetID[asgIDLowercase]; !found {
 			continue
 		}
--- a/pkg/cloud-provider/cloudapi/azure/azure_security.go
+++ b/pkg/cloud-provider/cloudapi/azure/azure_security.go
@@ -27,9 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	"antrea.io/nephe/apis/crd/v1alpha1"
-	"antrea.io/nephe/pkg/cloud-provider/cloudapi/common"
 	"antrea.io/nephe/pkg/cloud-provider/securitygroup"
-	"antrea.io/nephe/pkg/cloud-provider/utils"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkg/controllers/cloud/networkpolicy_sync.go
```diff
@@ -290,7 +290,7 @@ func (r *NetworkPolicyReconciler) getNICsOfCloudResources(resources []*securityg
 	for _, rsc := range resources {
 		name := rsc.Name
```
reachjainrahul, line 291: is it name or ID. verify it for azure

pkg/cloud-provider/cloudapi/azure/azure_security.go
```diff
@@ -576,7 +573,7 @@ func (computeCfg *computeServiceConfig) processAndBuildATSgView(networkInterface
 		}
 		vnetIDLowerCase := strings.ToLower(*networkInterface.VnetID)
 		nsgIDToVnetIDMap[nsgIDLowerCase] = vnetIDLowerCase
-		if strings.Compare(sgName, strings.ToLower(appliedToSecurityGroupNamePerVnet)) == 0 {
+		if strings.Contains(strings.ToLower(sgName), appliedToSecurityGroupNamePerVnet) {
```
reachjainrahul, line 576: Can you create 2 VNET and verify if things look Ok there too.

pkg/cloud-provider/cloudapi/azure/azure_security.go
```diff
@@ -668,6 +662,10 @@ func (computeCfg *computeServiceConfig) getAGGroupView(nepheControllerAGSgNameTo
 	var enforcedSecurityCloudView []securitygroup.SynchronizationContent
 	for _, appSecurityGroup := range appSecurityGroups {
 		asgIDLowercase := strings.ToLower(*appSecurityGroup.ID)
+		// if Asg is not associated with a managed VPC, do not create sync content for it.
```
reachjainrahul, line 665: nit: ASG

pkg/cloud-provider/cloudapi/azure/azure_security.go
```diff
@@ -650,16 +646,14 @@ func (computeCfg *computeServiceConfig) getATGroupView(nepheControllerATSGNameTo
...
-
-			nepheControllerATSgNameSet[atSgName] = struct{}{}
 		}
 	}
 
-	return enforcedSecurityCloudView, nepheControllerATSgNameSet, nil
+	return enforcedSecurityCloudView, nil
 }
 
 func (computeCfg *computeServiceConfig) getAGGroupView(nepheControllerAGSgNameToCloudResourcesMap map[string][]securitygroup.CloudResource,
```
reachjainrahul, line 655: nit: please add comment as and when you touch some code which doesnt have it.

pkg/controllers/cloud/networkpolicy_sync.go
```diff
@@ -28,59 +28,59 @@ import (
...
+		sg.state = securityGroupStateCreated
+		syncMembers := make([]*securitygroup.CloudResource, 0, len(syncContent.Members))
+		for i := range syncContent.Members {
+			syncMembers = append(syncMembers, &syncContent.Members[i])
 		}
 
-		nMembers := s.members
-		if len(cMembers) > 0 && cMembers[0].Type == securitygroup.CloudResourceTypeNIC {
-			nMembers, _ = r.getNICsOfCloudResources(s.members)
+		internalMembers := sg.members
```
reachjainrahul, line 41: nit: cachedMembers?

pkg/cloud-provider/cloudapi/azure/azure_security.go
archanapholla: Yes, I have done it. Both the VNETs had an ANP and I could see sync happening on both.
Logically speaking, this code is just checking if the NSG has specific format. A few lines above this code, we get NSGs associated with network interface of managed VPCs. So, at this block every NSG is checked against the condition and goes inside the if condition.

pkg/cloud-provider/cloudapi/azure/azure_security.go
archanapholla: added.

pkg/controllers/cloud/networkpolicy_sync.go
archanapholla: yes, probably more appropriate.

pkg/controllers/cloud/networkpolicy_sync.go
```diff
@@ -28,59 +28,59 @@ import (
...
 
-	if s.state == securityGroupStateCreated {
-		log.V(1).Info("Update securityGroup", "Name", s.id.Name, "MembershipOnly", membershipOnly, "CloudSecurityGroup", c)
-		_ = s.updateImpl(csg, nil, nil, membershipOnly, r)
-	} else if s.state == securityGroupStateInit {
-		log.V(1).Info("Add securityGroup", "Name", s.id.Name, "MembershipOnly", membershipOnly, "CloudSecurityGroup", c)
-		_ = s.addImpl(csg, membershipOnly, r)
+	if sg.state == securityGroupStateCreated {
+		log.V(1).Info("Update securityGroup", "Name", sg.id.Name, "MembershipOnly", membershipOnly, "CloudSecurityGroup", syncContent)
+		_ = sg.updateImpl(csg, nil, nil, membershipOnly, r)
```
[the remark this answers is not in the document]
archanapholla: Added, printing both cached SG members and sync members, this should help figure out why member compare failed.

pkg/cloud-provider/cloudapi/azure/azure_security.go
```diff
@@ -650,16 +649,15 @@ func (computeCfg *computeServiceConfig) getATGroupView(nepheControllerATSGNameTo
...
-
-			nepheControllerATSgNameSet[atSgName] = struct{}{}
 		}
 	}
 
-	return enforcedSecurityCloudView, nepheControllerATSgNameSet, nil
+	return enforcedSecurityCloudView, nil
 }
 
+// getAGGroupView creates synchronization content for ASGs created by nephe under managed VPCs.
```
reachjainrahul, line 658: nit: VNET?

pkg/controllers/cloud/networkpolicy_sync.go
```diff
@@ -28,59 +28,64 @@ import (
...
-		if len(cMembers) > 0 && cMembers[0].Type == securitygroup.CloudResourceTypeNIC {
-			nMembers, _ = r.getNICsOfCloudResources(s.members)
+		cachedMembers := sg.members
+		if len(syncMembers) > 0 && syncMembers[0].Type == securitygroup.CloudResourceTypeNIC {
+			cachedMembers, _ = r.getNICsOfCloudResources(sg.members)
 		}
-		if compareCloudResources(nMembers, cMembers) {
-			log.V(1).Info("Same SecurityGroup found", "Name", s.id.Name, "State", s.state)
+		if compareCloudResources(cachedMembers, syncMembers) {
+			log.V(1).Info("Everything in sync with cloud", "Name", sg.id.Name, "State", sg.state)
```
reachjainrahul, line 46: actually I want logging to be reduced.. This will again show that member is synced.. At every interval the minimal logging I am looking for is
1) Syncing cloud entities.. 
2) Figure out this entity requires re-sync and just dump that.

With that, I would implicitly know those entities are not in sync, rather than seeing success logs for all.

reachjainrahul:
/nephe-test-e2e-kind, /nephe-test-e2e-aks, /nephe-test-e2e-eks

pkg/cloud-provider/cloudapi/azure/azure_security.go
archanapholla: Done

pkg/cloud-provider/cloudapi/azure/azure_security.go
archanapholla: Done

pkg/controllers/cloud/networkpolicy_sync.go
archanapholla: Removed the logging when members are in sync. Otherwise, added a log.

pkg/controllers/cloud/networkpolicy_sync.go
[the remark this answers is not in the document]
reachjainrahul: I think this will be logged right??

PR #92, opened 2023-01-13
Disable vpc inventory test

Anandkumar26:
- The expected output and the actual output needs to be sorted before comparison.


reachjainrahul (approved):
/LGTM

PR #96, opened 2023-01-16
Pin K8s to 1.23.12 for AKS

reachjainrahul:
K8s 1.24 version introduced breaking change on how Service Accounts (SA) are handled. Pinning the K8s version till Antrea updates SA for VM agent support.


Anandkumar26 (approved):
/LGTM

PR #100, opened 2023-01-18
Fix for creating synchronization content for AT SG.

archanapholla:
- An example of NSG in Azure is nephe-at-per-vnet-default-archana-test-vnet-1. Fix is related to string comparison of SG name with "per-vnet-default". As this is not a direct comparison use "Contains" instead of "Compare".


archanapholla:
Change is taken care in https://github.com/antrea-io/nephe/pull/90 PR

PR #101, opened 2023-01-18
Fix to create sync content for AG when AT with same name exists.

archanapholla:
- Irrespective of AT group names, fix creates sync content for AG.
Name collision between AG and AT group can arise when two ANP's with vm1 AT -> to vm2 AG and vm2 AT -> vm1 AG policy respectively.


archanapholla:
Change is taken care in https://github.com/antrea-io/nephe/pull/90 PR

PR #102, opened 2023-01-19
Propagate VPC poll status to the caller for updating CPA Status field.

archanapholla:
If fetching vpc list from cloud returns an error, we continue to fetch instances from cloud with the intention that fetching vm instances has a higher priority over fetching vpc list. However, status of vpc poll needs to be taken into consideration and use it for setting "Error" in Status field of CPA CR. This plays a major role especially when CPA is added with incorrect credentials and CES is not yet added.

With this change, "Error" in Status field of CPA CR gets immediately set when fetching vpcs from cloud returns an error.


reachjainrahul (approved):
/LGTM

PR #105, opened 2023-01-20
Remove unnecessary namespace tests

reachjainrahul:
All ANPs were applied to same NS. Hence removing tests for diff NS. In future when we will create resources in 2 Namespaces, diff NS tests will be re-added.


Anandkumar26 (approved):
/LGTM

PR #114, opened 2023-02-03
Fix log and error messages

Anandkumar26:
Info logs should always start with capital letter.
Error logs should always start with small letter.
    
Also add sleep in crd integration test, to allow CR create/delete operation to be processed by the reconciler
Also handle some error cases in the account poller effectively.
    
### DIFF
--- a/pkg/controllers/cloud/account_poller.go
+++ b/pkg/controllers/cloud/account_poller.go
@@ -78,7 +78,7 @@ func (p *Poller) addAccountPoller(cloudType cloudv1alpha1.CloudProvider, namespa
 	defer p.mutex.Unlock()
 
 	if pollerScope, exists := p.accPollers[*namespacedName]; exists {
-		p.log.Info("poller exists", "account", namespacedName)
+		p.log.Info("Poller exists", "account", namespacedName)
 		return pollerScope, exists
 	}
 
@@ -137,7 +137,7 @@ func (p *Poller) addAccountPoller(cloudType cloudv1alpha1.CloudProvider, namespa
 		})
 
 	p.accPollers[*namespacedName] = poller
-	p.log.Info("poller will be created", "account", namespacedName)
+	p.log.Info("Poller will be created", "account", namespacedName)
 	return poller, false
 }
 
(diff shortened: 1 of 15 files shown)


pkg/controllers/cloud/networkpolicy_sync.go
```diff
@@ -255,6 +255,7 @@ func (r *NetworkPolicyReconciler) syncWithCloud() {
 	log := r.Log.WithName("CloudSync")
 
 	if r.bookmarkCnt < npSyncReadyBookMarkCnt {
+		log.V(1).Info("Skipping sync wait for bookmark events to be processed")
```
reachjainrahul, line 258: Dont add this, otherwise on restart for each ANP, it will be logged

pkg/controllers/cloud/virtualmachine_controller.go
```diff
@@ -50,7 +50,7 @@ func (r *VirtualMachineReconciler) Reconcile(_ context.Context, req ctrl.Request
 			return ctrl.Result{}, err
 		}
 		// Fall through if owner is deleted.
-		r.Log.V(1).Info("Is delete", "vm", req.NamespacedName)
+		r.Log.V(1).Info("Is VM crd deleted", "vm", req.NamespacedName)
```
reachjainrahul, line 53: Why is this 'Is'?

pkg/converter/source/virtualmachine_converter.go
```diff
@@ -62,22 +62,22 @@ func (v VMConverter) Start() {
...
-	log := v.Log.WithName("processEvent")
 
-	if isAgent {
+	if isExternalnode {
 		fetchKey = target.GetExternalNodeKeyFromSource(vm)
 	} else {
 		fetchKey = target.GetExternalEntityKeyFromSource(vm)
 	}
-	log.Info("Received event", "Key", fetchKey, "Agented", isAgent)
+	v.Log.Info("Received event", "Key", fetchKey, "Agented", isExternalnode)
```
reachjainrahul, line 74: Why Agented here, and isExternalNode later.. be consistent.. Probably just use ExternalNode?

pkg/controllers/cloud/networkpolicy_sync.go
Anandkumar26: Done

pkg/controllers/cloud/virtualmachine_controller.go
Anandkumar26: Updated the logging

pkg/controllers/cloud/networkpolicy_sync.go
```diff
@@ -291,6 +291,7 @@ func (r *NetworkPolicyReconciler) syncWithCloud() {
 			}
 		}
 	}
+	log.V(1).Info("Synchronization content is nil")
```
archanapholla, line 294: @Anandkumar26 as discussed offline, could you please add 2 info logs to figure out start and end of cloud sync. This is to understand from logs the course of action taken place in cloud sync. 
Also, in line 278, could you please explicitly state that SG is not present in cache, hence deleting it?

pkg/controllers/cloud/networkpolicy_sync.go
```diff
@@ -261,6 +261,7 @@ func (r *NetworkPolicyReconciler) syncWithCloud() {
 	if r.bookmarkCnt < npSyncReadyBookMarkCnt {
 		return
 	}
+	log.V(1).Info("Started synchronizing security groups in controller with cloud")
```
reachjainrahul, line 264: Started synchronizing cloud security groups with controller
```suggestion
	log.V(1).Info("Started synchronizing security groups in controller with cloud")
```

reachjainrahul, line 264:
```suggestion
	log.V(1).Info("Started synchronizing security groups in controller with cloud")
	log.V(1).Info("Started synchronizing cloud security groups with controller")
```

reachjainrahul, line 264:
```suggestion
	log.V(1).Info("Started synchronizing cloud security groups with controller")
```

pkg/controllers/cloud/networkpolicy_sync.go
```diff
@@ -317,6 +319,7 @@ func (r *NetworkPolicyReconciler) syncWithCloud() {
 			break
 		}
 	}
+	log.V(1).Info("Done synchronizing security groups in controller with cloud")
```
reachjainrahul, line 322: Same here

pkg/controllers/cloud/account_poller.go
```diff
@@ -243,20 +243,20 @@ func (p *accountPoller) doAccountPoller() {
...
+		p.log.Error(e, "failed to fetch cloud vpc list from internal snapshot", "account",
+			p.namespacedName.String())
 	}
 
-	err := p.inventory.BuildVpcCache(vpcMap, p.namespacedName)
-	if err != nil {
-		p.log.Info("failed to build vpc inventory", "account", p.namespacedName.String(), "error", err)
+	e = p.inventory.BuildVpcCache(vpcMap, p.namespacedName)
+	if e != nil {
+		p.log.Error(e, "failed to build vpc cache", "account", p.namespacedName.String())
```
reachjainrahul, line 259: logic point of view, if there is error is building vpc cache, is there any point continuing further?

Anandkumar26: p.inventory.BuildVpcCache - returns an error only when creating or updating a VPC..
When it fails, we should still continue to pull VM inventory, because VM resources might have changed.


vpcMap, e := cloudInterface.GetVpcInventory(p.namespacedName) - returns an error when configs are not present in cache. It implies CPA config is invalid. We should not continue forward and return when an error occurs.

reachjainrahul (approved):
/LGTM

PR #116, opened 2023-02-08
Add endpoint option in cloudprovideraccount CRD

reachjainrahul:
Endpoint can be used to set non AWS default URL. Primarily used to set endpoint to mock AWS services like localstack


Anandkumar26 (approved):
/LGTM

PR #118, opened 2023-02-08
Move VpcInfo from Spec to Status

reachjainrahul:
VpcInfo is discovered from cloud. Hence moving the info from spec to status, similar to virtual machine CR.


Anandkumar26 (approved):
/LGTM

PR #119, opened 2023-02-09
Send rule realization update even if ANP didnt change

reachjainrahul:
If an ANP is configured with Allow -> Reject -> Allow actions, nephe was not sending rule realization update as it didnt find any diff in the rules. As Reject action is failed early before internal cache in nephe is updated. Fix is to send ANP realization update even if no change in ANP rules are detected.
This will handle antrea controller restart case too, in which when all ANPs are replayed by antrea controller, nephe will not detect any change and no rule status update will happen.


Anandkumar26 (approved):
/LGTM

PR #121, opened 2023-02-09
Sync with cloud feature hardening - 2.

archanapholla:
Fixes:
- Add empty string check for sg rule description field while creating sync content.
- Stale AT SG doesn't get deleted during sync as sync doesn't identify it as an ASG associated with NSG. It is due to the usage of ASGs in "tags" field of NSG for retrieving ASGs from cloud. User added stale ASG can't be present in tags. Fix is to use ASGs attached to network interfaces instead of getting ASGs from "tags". AT SGs are added as "tags" in a NSG for user visibility purpose.
- SG deletion logic uses RG name from Vnet of CloudResource. If SG belongs to a different RG then it SG doesn't get deleted. Hence, while creating sync content make sure that only SGs with RGs same as Vnet are included.

### DIFF
--- a/pkg/cloud-provider/cloudapi/azure/azure_security.go
+++ b/pkg/cloud-provider/cloudapi/azure/azure_security.go
@@ -84,7 +84,7 @@ func (computeCfg *computeServiceConfig) processAppliedToMembership(appliedToGrou
 
 	// find network interfaces which are using or need to use the provided NSG
 	nwIntfIDSetNsgToAttach := make(map[string]struct{})
-	nwIntfIDSetNsgToDettach := make(map[string]struct{})
+	nwIntfIDSetNsgToDetach := make(map[string]struct{})
 	for _, networkInterface := range networkInterfaces {
 		nwIntfIDLowerCase := strings.ToLower(*networkInterface.ID)
 		// 	for network interfaces not attached to any virtual machines, skip processing
@@ -101,11 +101,22 @@ func (computeCfg *computeServiceConfig) processAppliedToMembership(appliedToGrou
 				azurePluginLogger().Error(err, "nsg ID format not valid", "nsgID", nsgID)
 				return err
 			}
-			if len(networkInterface.Tags) > 0 {
-				tags := networkInterface.Tags[0]
-				_, found := tags[cloudSgNameLowercase]
-				if strings.Compare(nsgNameLowercase, perVnetNsgNameLowercase) == 0 && found {
-					isNsgAttached = true
+
+			attachedAsgIDs := networkInterface.ApplicationSecurityGroupIDs
+			for _, asgID := range attachedAsgIDs {
+				asgIDLowerCase := strings.ToLower(*asgID)
+				// proceed only if network-interface attached to nephe ASG
+				_, _, asgName, err := extractFieldsFromAzureResourceID(asgIDLowerCase)
+				if err != nil {
+					continue
+				}
+				_, _, isAT := securitygroup.IsNepheControllerCreatedSG(asgName)
+				if isAT {
+					if strings.Compare(nsgNameLowercase, perVnetNsgNameLowercase) == 0 &&
+						strings.Compare(cloudSgNameLowercase, asgName) == 0 {
+						isNsgAttached = true
+						break
+					}
 				}
 			}
 		}
@@ -113,7 +124,7 @@ func (computeCfg *computeServiceConfig) processAppliedToMembership(appliedToGrou
 		_, isNicMemberNetworkInterface := memberNetworkInterfaces[strings.ToLower(*networkInterface.ID)]
 		if isNsgAttached {
 			if !isNicAttachedToMemberVM && !isNicMemberNetworkInterface {
-				nwIntfIDSetNsgToDettach[nwIntfIDLowerCase] = struct{}{}
+				nwIntfIDSetNsgToDetach[nwIntfIDLowerCase] = struct{}{}
 			}
 		} else {
 			if isNicAttachedToMemberVM || isNicMemberNetworkInterface {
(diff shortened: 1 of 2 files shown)


pkg/cloud-provider/cloudapi/azure/azure_security.go
```diff
@@ -221,6 +234,7 @@ func (computeCfg *computeServiceConfig) processAddressGroupMembership(addressGro
 	return computeCfg.processAsgAttachDetachConcurrently(asgObj, nwIntfIDSetAsgToAttach, nwIntfIDSetAsgToDettach)
 }
 
+// processAsgAttachDetachConcurrently attaches/detaches network interfaces to/from an ASG.
```
reachjainrahul, line 237: ASG or NSG?

pkg/cloud-provider/cloudapi/azure/azure_security.go
```diff
@@ -550,7 +566,34 @@ func (computeCfg *computeServiceConfig) processAndBuildAGSgView(
 	return addressGroupSgEnforcedView, err
 }
 
-// processAndBuildAGSgView creates synchronization content for AdressGroup SG.
+// isValidAtSg finds if an ASG is a valid nephe created AppliedTo SG and belong to same RG as VNET.
+func (computeCfg *computeServiceConfig) isValidAtSg(asgID string, vnetID string) (string, error) {
```
reachjainrahul, line 570: use AT in caps, or isValidAppliedToSG.. I would prefer later

pkg/cloud-provider/cloudapi/azure/azure_security.go
```diff
@@ -655,6 +703,40 @@ func (computeCfg *computeServiceConfig) getATGroupView(nepheControllerATSGNameTo
 	return enforcedSecurityCloudView, nil
 }
 
+// isValidAgSg finds if an ASG is in managed VNET, and it is a valid nephe created AddressGroup SG.
+func (computeCfg *computeServiceConfig) isValidAgSg(asgIDLowercase string,
```
reachjainrahul, line 707: same here.. also see if you can move functions together. isValidAG and AT onces . back to back..

pkg/cloud-provider/cloudapi/azure/azure_security.go
```diff
@@ -655,6 +703,40 @@ func (computeCfg *computeServiceConfig) getATGroupView(nepheControllerATSGNameTo
...
+
+	// proceed only if ASG is a nephe created AddressGroup SG.
+	_, asgRgName, cloudAsgName, err := extractFieldsFromAzureResourceID(asgIDLowercase)
+	if err != nil {
+		return "", "", err
+	}
+
+	asgName, isAG, _ := securitygroup.IsNepheControllerCreatedSG(cloudAsgName)
+	if !isAG {
+		return "", "", fmt.Errorf("asg is not a nephe created AddressGroup SG [%v]", cloudAsgName)
```
reachjainrahul, line 723: why you need to throw error if asg is not nephe created.. this function should just check if asg is nephe created or not

pkg/cloud-provider/cloudapi/azure/azure_security.go
archanapholla: This is an ASG, we have a separate function for NSG(processNsgAttachDetachConcurrently) used for AT group membership update.

pkg/cloud-provider/cloudapi/azure/azure_security.go
archanapholla: I didn't want to club them together as both the functions have different signatures.

pkg/cloud-provider/cloudapi/azure/azure_security.go
archanapholla: This error is only for indication purpose on the caller side to figure out if CloudResource(used for creating sync content at later stage) should be created for the ASG or not . Not really flagging it anywhere.

pkg/cloud-provider/cloudapi/azure/azure_security.go
archanapholla: Yes done

pkg/cloud-provider/cloudapi/azure/azure_security.go
archanapholla: Added bool for the return value as discussed.

pkg/cloud-provider/cloudapi/azure/azure_security.go
archanapholla: Re-checked, new functions are placed right, re-arranged an existing function to club all AT SG related functions together, similarly AG.

pkg/cloud-provider/cloudapi/azure/azure_security.go
```diff
@@ -655,6 +660,85 @@ func (computeCfg *computeServiceConfig) getATGroupView(nepheControllerATSGNameTo
 	return enforcedSecurityCloudView, nil
 }
 
+// isValidAddressGroupSg finds if an ASG is in managed VNET, and it is a valid nephe created AddressGroup SG.
+func (computeCfg *computeServiceConfig) isValidAddressGroupSg(asgIDLowercase string,
```
reachjainrahul, line 664: Can you merge isValidAddrGroup and isValidAppliedTo. Both share same logic.. Use bool to indicate if check if for AG or AT

archanapholla: I purposely kept them separate as there is an extra operation in isValidAdressGroupSg() to figure out if asg is associated with network interface of managed vnet.  This is not needed in the other. Due to which function signatures are also different.
// proceed only if ASG is associated with network interfaces of managed VNET.
	vnetID, found := asgIDToVnetID[asgIDLowercase]
	if !found {
		return "", "", false
	}

pkg/cloud-provider/cloudapi/azure/azure_security.go
```diff
@@ -506,51 +522,35 @@ func getAsgsToAdd(asgs *[]network.ApplicationSecurityGroup, addrGroupNepheContro
...
+	// proceed only if RG of ASG is same as VNET RG(deletion of ASG uses RG from VNET ID).
+	_, vnetRg, _, err := extractFieldsFromAzureResourceID(vnetID)
+	if err != nil {
+		return "", false
 	}
 
-	addressGroupSgEnforcedView, err := computeCfg.getAGGroupView(nepheControllerAGSgNameToMemberCloudResourcesMap, asgIDToVnetIDMap)
-	return addressGroupSgEnforcedView, err
+	if strings.Compare(asgRgName, vnetRg) != 0 {
+		azurePluginLogger().V(1).Info("RG of ASG is not same as RG of VNET", "ASG", asgID, "ASG RG",
```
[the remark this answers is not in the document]
archanapholla: Its printed if we have ASGs attached to our instances which do not belong to the same vnet. If you think its too chatty I am totally fine with removing it, just added it for debugging purpose.

reachjainrahul (approved):
/LGTM

PR #123, opened 2023-02-13
Add Managed VPC flag in VPC Info

reachjainrahul:
Managed VPCs/VNETs are those which include one ore more managed VMs. Added a flag in VPC info struct to indicate if the VPC is managed or not.
Also updated get vpc output and doc.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -195,22 +195,22 @@ kubectl get vpc -A
 
 ```text
 # Output
-NAMESPACE   NAME                    CLOUD PROVIDER   REGION
-sample-ns   vpc-0f54c9f1b395038ab   AWS              us-west-1
-sample-ns   vpc-04269a331ab6cd649   AWS              us-west-1
-sample-ns   vpc-047156bebab1083c9   AWS              us-west-1
+NAMESPACE   NAME                    CLOUD PROVIDER   REGION      MANAGED
+sample-ns   vpc-0d6bb6a4a880bd9ad   AWS              us-west-1   true
+sample-ns   vpc-04269a331ab6cd649   AWS              us-west-1   false
+sample-ns   vpc-047156bebab1083c9   AWS              us-west-1   false
 ```
 
 Use describe on VPC object to get `Id` or `Name` field and use it in vpcMatch
 section of `CloudEntitySelector` configuration.
 
 ```bash
-kubectl describe vpc vpc-0f54c9f1b395038ab -n sample-ns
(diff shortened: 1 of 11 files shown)


reachjainrahul:
/nephe-test-e2e-kind, /nephe-test-e2e-aks, /nephe-test-e2e-eks

docs/getting-started.md
```diff
@@ -210,7 +210,7 @@ kubectl describe vpc vpc-0f54c9f1b395038ab -n sample-ns
 
 ```text
 # Output
-Name:         vpc-0f54c9f1b395038ab
+Name:         vpc-0d6bb6a4a880bd9ad
```
archanapholla, line 213: Also, can we update line 208 to have vpc-0d6bb6a4a880bd9ad instead of vpc-0f54c9f1b395038ab?

pkg/cloud-provider/cloudapi/azure/azure_compute.go
```diff
@@ -348,11 +349,17 @@ func (computeCfg *computeServiceConfig) GetVpcInventory() map[string]*runtimev1a
...
 
+	vnetIDs := computeCfg.getManagedVnetIDs()
+
 	// Convert to kubernetes object and return a map indexed using VnetID.
 	vpcMap := map[string]*runtimev1alpha1.Vpc{}
 	for _, vpc := range snapshot.(*computeResourcesCacheSnapshot).vnets {
 		if strings.EqualFold(*vpc.Location, computeCfg.credentials.region) {
-			vpcObj := ComputeVpcToInternalVpcObject(&vpc, tokens[0], tokens[1], strings.ToLower(computeCfg.credentials.region))
+			managed := false
+			if _, ok := vnetIDs[strings.ToLower(*vpc.ID)]; ok {
```
archanapholla, line 359: It looks like when we add vnetIDs to snapshot (in DoResourceInventory function), we do not convert vpc.ID to lowercase, hence I think we should avoid conversion to lower case while fetching from vnetIDs map here.

docs/getting-started.md
reachjainrahul: Done

pkg/cloud-provider/cloudapi/azure/azure_compute.go
reachjainrahul: we are fetching id from vm struct in doRInventory and that is in lower case.. so lower case is required here.

archanapholla: Yes right, I verified it.

archanapholla (approved):
LGTM

PR #132, opened 2023-02-21
Update description field in cloud security rule

Anandkumar26:
Each cloud security rule created by nephe will have ANP name, ANP namespace and an AppliedToGroup field.

Ignore any rules which don't have a description field or  have an invalid description field.
    
### DIFF
--- a/pkg/cloud-provider/cloudapi/aws/aws_converters.go
+++ b/pkg/cloud-provider/cloudapi/aws/aws_converters.go
@@ -44,11 +44,12 @@ func convertToIPPermissionPort(port *int, protocol *int) (*int64, *int64) {
 	return portVal, portVal
 }
 
-func convertToEc2IpRanges(ips []*net.IPNet, ruleHasGroups bool) []*ec2.IpRange {
+func convertToEc2IpRanges(ips []*net.IPNet, ruleHasGroups bool, description *string) []*ec2.IpRange {
 	var ipRanges []*ec2.IpRange
 	if len(ips) == 0 && !ruleHasGroups {
 		ipRange := &ec2.IpRange{
-			CidrIp: aws.String("0.0.0.0/0"),
+			CidrIp:      aws.String("0.0.0.0/0"),
+			Description: description,
 		}
 		ipRanges = append(ipRanges, ipRange)
 		return ipRanges
@@ -56,28 +57,33 @@ func convertToEc2IpRanges(ips []*net.IPNet, ruleHasGroups bool) []*ec2.IpRange {
 
 	for _, ip := range ips {
 		ipRange := &ec2.IpRange{
-			CidrIp: aws.String(ip.String()),
+			CidrIp:      aws.String(ip.String()),
+			Description: description,
 		}
 		ipRanges = append(ipRanges, ipRange)
 	}
 	return ipRanges
 }
 
-func convertFromIPRange(ipRanges []*ec2.IpRange) []*net.IPNet {
+func convertFromIPRange(ipRanges []*ec2.IpRange) ([]*net.IPNet, []*string) {
 	var srcIPNets []*net.IPNet
+	var desc []*string
 	for _, ipRange := range ipRanges {
 		_, ipNet, err := net.ParseCIDR(*ipRange.CidrIp)
 		if err != nil {
 			continue
 		}
+		desc = append(desc, ipRange.Description)
 		srcIPNets = append(srcIPNets, ipNet)
 	}
-	return srcIPNets
+
+	return srcIPNets, desc
 }
 
 func convertFromSecurityGroupPair(cloudGroups []*ec2.UserIdGroupPair, managedSGs map[string]*ec2.SecurityGroup,
-	unmanagedSGs map[string]*ec2.SecurityGroup) []*securitygroup.CloudResourceID {
+	unmanagedSGs map[string]*ec2.SecurityGroup) ([]*securitygroup.CloudResourceID, []*string) {
 	var cloudResourceIDs []*securitygroup.CloudResourceID
+	var desc []*string
 
 	for _, cloudGroup := range cloudGroups {
 		var sgName string
@@ -87,11 +93,13 @@ func convertFromSecurityGroupPair(cloudGroups []*ec2.UserIdGroupPair, managedSGs
 		if foundInManagedSg {
 			sgName, _, _ = securitygroup.IsNepheControllerCreatedSG(*managedSgObj.GroupName)
 			vpcID = *managedSgObj.VpcId
+			desc = append(desc, cloudGroup.Description)
 		}
 		unmanagedSGObj, foundInUnmanagedSg := unmanagedSGs[*cloudGroup.GroupId]
 		if foundInUnmanagedSg {
 			sgName = *unmanagedSGObj.GroupName
 			vpcID = *unmanagedSGObj.VpcId
+			desc = append(desc, cloudGroup.Description)
 		}
 
 		cloudResourceID := &securitygroup.CloudResourceID{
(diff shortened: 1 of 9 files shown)


pkg/cloud-provider/cloudapi/aws/aws_converters.go
```diff
@@ -101,16 +109,24 @@ func convertFromSecurityGroupPair(cloudGroups []*ec2.UserIdGroupPair, managedSGs
...
 	for _, ipPermission := range ipPermissions {
-		fromSrcIPs := convertFromIPRange(ipPermission.IpRanges)
-		for _, srcIP := range fromSrcIPs {
+		fromSrcIPs, desc := convertFromIPRange(ipPermission.IpRanges)
+		for i, srcIP := range fromSrcIPs {
+			// Get cloud rule description.
+			_, err := securitygroup.ExtractCloudDescription(desc[i])
+			if err != nil {
+				// Ignore rules that don't have a valid description field.
+				awsPluginLogger().Error(err, "failed to extract cloud rule description", "desc", desc[i])
```
reachjainrahul, line 127: log as debug, and will it be printed on every sync interval

pkg/cloud-provider/cloudapi/aws/aws_security.go
```diff
@@ -204,8 +205,12 @@ func (ec2Cfg *ec2ServiceConfig) realizeIngressIPPermissions(cloudSgObj *ec2.Secu
 		if rule == nil {
 			continue
 		}
-		idGroupPairs := buildEc2UserIDGroupPairs(rule.FromSecurityGroups, cloudSGNameToObj)
-		ipRanges := convertToEc2IpRanges(rule.FromSrcIP, len(rule.FromSecurityGroups) > 0)
+		description, err := securitygroup.GenerateCloudDescription(obj.NetworkPolicy, *cloudSgObj.GroupName)
+		if err != nil {
+			return fmt.Errorf("unable to generate rule description, err: %v", err)
```
reachjainrahul, line 210: this shouldnt happen right for all practical purpose?

pkg/cloud-provider/cloudapi/aws/aws_security_test.go
```diff
@@ -323,9 +359,100 @@ var _ = Describe("AWS Cloud Security", func() {
 			err := cloudInterface.UpdateSecurityGroupRules(webSgIdentifier, addRule, []*securitygroup.CloudRule{}, addRule)
 			Expect(err).Should(BeNil())
 		})
+		It("Should fail to create egress rules", func() {
```
reachjainrahul, line 362: what is generating fail case?? add a comment there too

pkg/cloud-provider/cloudapi/aws/aws_converters.go
Anandkumar26: Ok will do.

pkg/cloud-provider/cloudapi/aws/aws_security.go
Anandkumar26: Yes.

pkg/cloud-provider/cloudapi/aws/aws_security_test.go
Anandkumar26: Add a comment explaining the test scenario.

pkg/cloud-provider/cloudapi/aws/aws_converters.go
Anandkumar26: Changed the log level.

pkg/cloud-provider/cloudapi/aws/aws_security_test.go
Anandkumar26: Done

reachjainrahul (approved):
/LGTM

pkg/cloud-provider/securitygroup/helpers.go
```diff
@@ -15,9 +15,12 @@
 package securitygroup
 
 import (
+	"fmt"
 	"strings"
 )
 
+// IsNepheControllerCreatedSG checks an SG is created by nephe
+// and returns if it's an AppliedToGroup or AddressGroup Sg and Sg name
```
shenmo3, line 23: nit: SG or sg, also period at the end.

pkg/cloud-provider/securitygroup/securitygroup.go
```diff
@@ -170,6 +177,18 @@ func (c *CloudResourceID) String() string {
 	return c.Name + "/" + c.Vpc
 }
 
+type CloudRuleDescription struct {
+	Name           string
+	Namespace      string
+	AppliedToGroup string
+}
+
+func (r *CloudRuleDescription) String() string {
```
shenmo3, line 186: Maybe we should split all the cloud resource related definitions and funcs into a separate file, instead of reusing the securitygroup.go?

pkg/cloud-provider/securitygroup/securitygroup.go
```diff
@@ -170,6 +177,18 @@ func (c *CloudResourceID) String() string {
 	return c.Name + "/" + c.Vpc
 }
 
+type CloudRuleDescription struct {
+	Name           string
+	Namespace      string
```
shenmo3, line 182: Should we say ANPName and ANPNamespace, or mention ANP in a comment?

pkg/cloud-provider/securitygroup/securitygroup.go
```diff
@@ -170,6 +177,18 @@ func (c *CloudResourceID) String() string {
...
+	AppliedToGroup string
+}
+
+func (r *CloudRuleDescription) String() string {
+	return Name + ":" + r.Name + ", " +
+		Namespace + ":" + r.Namespace + ", " +
+		AppliedToGroup + ":" + r.AppliedToGroup
```
shenmo3, line 189: Use consts for `:`, `,` separators?

pkg/cloud-provider/cloudapi/aws/aws_converters.go
```diff
@@ -134,12 +157,20 @@ func convertFromIPPermissionToIngressRule(ipPermissions []*ec2.IpPermission, man
...
-		toDstIPs := convertFromIPRange(ipPermission.IpRanges)
-		for _, dstIP := range toDstIPs {
+		toDstIPs, desc := convertFromIPRange(ipPermission.IpRanges)
+		for i, dstIP := range toDstIPs {
+			// Get cloud rule description.
+			_, ok := securitygroup.ExtractCloudDescription(desc[i])
+			if !ok {
+				// Ignore rules that don't have a valid description field.
+				awsPluginLogger().V(4).Info("Failed to extract cloud rule description", "desc", desc[i])
+				continue
```
shenmo3, line 172: I believe this is used in cloudSync, so in this empty description case we should still converts the cloud user created rule, and report it to the cloudRule indexer. Otherwise the updateANPRules won't account for this user created rule and might lead to duplicate rules not caught.

Anandkumar26: This is an initial change to handle description field in cloud.
With this change intension is to find a way to distinguish user created rule vs nephe created rule.

Now coming to logical change, which will be done in a subsequent patch.
how do you envision handling user created vs nephe created rule? Based on that we can tweak what rules are to be stored in the synchronisation content.

shenmo3: Yes I understand, but here it would cause existing functionality to break. Currently we put all rules in sync content including user rules and add them into the cloudRule indexer. If we skip any rules here, update rule routine wouldn't be able to compute and claim the rule, lead to duplicate rule error from AWS, and unnecessary cloud api calls in Azure. We shouldn't skip any rules here, just return all rules and we can add identification of user rules in the next patch

Anandkumar26: Discussed the behavior and this will be addressed in a follow  up patch, as it is not under the scope of this PR.

pkg/cloud-provider/securitygroup/securitygroup.go
Anandkumar26: IMO, its not needed.

pkg/cloud-provider/securitygroup/securitygroup.go
Anandkumar26: Done

pkg/cloud-provider/securitygroup/helpers.go
Anandkumar26: Done

pkg/cloud-provider/cloudapi/aws/aws_security_test.go
```diff
@@ -44,6 +44,7 @@ var _ = Describe("AWS Cloud Security", func() {
 		testVMID02  = "i-0b194935df0d83eb8"
 
 		testAccountNamespacedName = &types.NamespacedName{Namespace: "namespace01", Name: "account01"}
+		testAnpNamespace          = &types.NamespacedName{Namespace: "test-anp-ns", Name: "test-anp"}
```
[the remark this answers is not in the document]
Anandkumar26: Done

pkg/cloud-provider/securitygroup/securitygroup.go
Anandkumar26: Since description is SG rule related, I prefer to keep it here.

reachjainrahul (approved):
/LGTM

PR #134, opened 2023-02-23
early exit on update rules when no rules to update

### LINKED ISSUE #124: Exit update rules early when there are no rules to update
**Describe the problem/challenge you have**
Currently the update rules routine enters the plugin level even when there are no rules to update.

**Describe the solution you'd like**
Exit the update rule routine at the network policy controller level after compute that there are no rules to update.

**Anything else you would like to add?**

shenmo3:
## Description
The current update rules routine proceeds to the cloud plugin level regardless of whether there are any actual updates. With this pull request, a check has been added to exit the routine early when there are no rules that require updating.

## Changes
1. Adds a check to exit updateANPRules func early when no rules to update.
2. Fixed a bug where security group deletion doesn't clean up remaining sg rules in cloudRule indexer.
3. Removed reconcile logic on security group retry success since it serves no purpose and leads to race condition.

Closes #124

### DIFF
--- a/pkg/controllers/cloud/networkpolicy.go
+++ b/pkg/controllers/cloud/networkpolicy.go
@@ -438,6 +438,9 @@ func (s *securityGroupImpl) deleteImpl(c cloudSecurityGroup, membershipOnly bool
 	ch := securitygroup.CloudSecurityGroup.DeleteSecurityGroup(&s.id, membershipOnly)
 	go func() {
 		err := <-ch
+		if err == nil && !membershipOnly {
+			s.deleteSgRulesFromIndexer(r)
+		}
 		r.cloudResponse <- &securityGroupStatus{sg: c, op: securityGroupOperationDelete, err: err}
 	}()
 	return nil
@@ -492,6 +495,18 @@ func (s *securityGroupImpl) notifyImpl(c PendingItem, membershipOnly bool, op se
 	}
 	if op == securityGroupOperationDelete {
 		_ = r.pendingDeleteGroups.Update(getGroupUniqueName(s.id.Name, membershipOnly), true, -1, !moreOps)
+	}
+}
+
+// deleteSgRulesFromIndexer deletes all rules that are part of the security group from the cloudRule indexer.
+func (s *securityGroupImpl) deleteSgRulesFromIndexer(r *NetworkPolicyReconciler) {
+	rules, err := r.cloudRuleIndexer.ByIndex(cloudRuleIndexerByAppliedToGrp, s.id.CloudResourceID.String())
+	if err != nil {
+		r.Log.Error(err, "get cloudRule indexer", "index", s.id.CloudResourceID.String())
+		return
+	}
+	for _, rule := range rules {
+		_ = r.cloudRuleIndexer.Delete(rule)
 	}
 }
 
@@ -684,7 +699,7 @@ func (a *addrSecurityGroup) removeStaleMembers(stales []*types.NamespacedName, r
 // appliedToSecurityGroup contains information to create a cloud appliedToSecurityGroup.
 type appliedToSecurityGroup struct {
 	securityGroupImpl
-	hasRules      bool
+	ruleReady     bool
 	hasMembers    bool
 	addrGroupRefs map[string]bool
 }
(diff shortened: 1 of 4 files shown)


pkg/controllers/cloud/networkpolicy.go
```diff
@@ -438,6 +439,14 @@ func (s *securityGroupImpl) deleteImpl(c cloudSecurityGroup, membershipOnly bool
 	ch := securitygroup.CloudSecurityGroup.DeleteSecurityGroup(&s.id, membershipOnly)
 	go func() {
 		err := <-ch
+		if err == nil && !membershipOnly {
+			// remove sg rules from the cache.
+			rules, e := r.cloudRuleIndexer.ByIndex(cloudRuleIndexerByAppliedToGrp, s.id.CloudResourceID.String())
+			if e != nil {
+				r.Log.Error(e, "get cloudRule indexer", "index", s.id.CloudResourceID.String())
+			}
+			_ = utils.DeleteIndexerObjects(r.cloudRuleIndexer, rules)
```
reachjainrahul, line 448: wont rules be nil, if error occurs ?

pkg/controllers/cloud/networkpolicy_pendingitem.go
```diff
@@ -245,20 +245,6 @@ func (p *pendingGroup) RunOrDeletePendingItem(id string, context interface{}) (r
 func (p *pendingGroup) ClearPendingState() {}
 
 func (s *securityGroupImpl) runPendingItemImpl(c cloudSecurityGroup, memberOnly bool, r *NetworkPolicyReconciler) bool {
-	if s.retryOp == nil {
-		// retry operation successful. Reconcile current group with
-		// plug-ing
-		if s.state != securityGroupStateGarbageCollectState {
```
reachjainrahul, line 251: why is this removed? can you explain the scenario?

pkg/controllers/utils/helpers.go
```diff
@@ -45,3 +48,43 @@ func GetAccountProviderType(account *v1alpha1.CloudProviderAccount) (v1alpha1.Cl
 		return "", fmt.Errorf("%s", ErrorMsgUnknownCloudProvider)
 	}
 }
+
+// UpdateIndexerObjects updates the objects in the indexer cache with the specified objects.
+func UpdateIndexerObjects(indexer cache.Indexer, objs interface{}) []error {
```
reachjainrahul, line 53: As you have moved this to helper, where is AddIndexerObject?? Also, dont think its generic helper.. You expect object to be added as slice, so its custom implementation. Find a place in networkpolicy for this

pkg/controllers/cloud/networkpolicy_pendingitem.go
shenmo3: Because its causing potential extra cloud calls and I don't see it serves any purpose at all. If a retry succeeded, the operation will send cloudResponse and the state machine continues, no reason to reconcile again here. 

In network policy unit test create fail case, this code is causing race conditions. On create fail, the operation is retried. After retry success, the state machine continues to update rules, and this reconcile code is also triggered, which also calls update rules.

pkg/controllers/cloud/networkpolicy.go
shenmo3: Yes, but nil slice is still slice, the function still works. We can add a check for nil to optimize a bit though.

pkg/controllers/utils/helpers.go
shenmo3: In indexer implementation, update and add are the same thing. In cloudRule indexer I'm also not using add at all, just update.

pkg/controllers/cloud/networkpolicy.go
```diff
@@ -762,7 +771,7 @@ func (a *appliedToSecurityGroup) updateAllRules(r *NetworkPolicyReconciler) erro
 
 // updateANPRules invokes cloud plug-in to update rules of appliedToSecurityGroup for a given ANP.
 func (a *appliedToSecurityGroup) updateANPRules(r *NetworkPolicyReconciler, np *networkPolicy) {
-	if !a.isReady() || a.deletePending || !np.rulesReady {
```
reachjainrahul, line 765: add a comment about each case

shenmo3: Done

pkg/controllers/cloud/networkpolicy.go
```diff
@@ -438,6 +438,9 @@ func (s *securityGroupImpl) deleteImpl(c cloudSecurityGroup, membershipOnly bool
 	ch := securitygroup.CloudSecurityGroup.DeleteSecurityGroup(&s.id, membershipOnly)
 	go func() {
 		err := <-ch
+		if err == nil && !membershipOnly {
+			s.deleteSgRulesFromIndexer(r)
```
reachjainrahul, line 442: we were not cleaning up cloudruleindexer earlier?

PR #140, opened 2023-03-01
Re-Add AWS plugin level lock

reachjainrahul:
Although AWS SDK is thread safe, the aws plugin caches in some state before calling the sdk, which leads to race condition. For example, if one interface is getting modified from two thread concurrently, the result is unpredictable as threads does get on previous state first modifies to new state and calls SDK. Due to this race condition, if thread1 removes SG1 and thread2 removes SG2, if thread2 is executed last, SG1 will not be removed and vice-versa.


Anandkumar26 (approved):
/LGTM

PR #142, opened 2023-03-02
Wait for ConfigMap controller to sync upon restart.

Anandkumar26:
ConfigMapController wil be first controller to process any CR. ConfigMapController is said to be synced only when nephe-config CR is received and processed by the reconciler.

Add ConfigMapController as dependent controller for CPA controller.


reachjainrahul (approved):
/LGTM

PR #144, opened 2023-03-03
Upgrade to kubernetes version 1.25 on AKS/EKS

Anandkumar26:
EKS nodes no longer use docker as container runtime. It uses containerd as container run time, similar to AKS

### DIFF
--- a/hack/eks
+++ b/hack/eks
@@ -20,7 +20,7 @@ TERRAFORM="terraform"
 KUBECTL="kubectl"
 RUN_PATH="$HOME/tmp/terraform-eks"
 CONFIG_PATH=$(dirname $(readlink -f "$0"))/terraform/eks
-CMD_ARRAY=($AWS_CLI $TERRAFORM $KUBECTL pv bzip2 jq)
+CMD_ARRAY=($AWS_CLI $TERRAFORM $KUBECTL pv jq)
 ENV_ARRAY=(TF_VAR_eks_cluster_iam_role_name TF_VAR_eks_iam_instance_profile_name TF_VAR_aws_key_pair_name)
 CREATE_ONLY=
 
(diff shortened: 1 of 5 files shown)


hack/eks
```diff
@@ -110,7 +110,7 @@ function load() {
   nodes=$(KUBECONFIG=$RUN_PATH/kubeconfig kubectl get nodes -o json | jq -r '.items[] | .status | .addresses[] | select(.type == "ExternalIP") |.address')
   for node in $(echo $nodes); do
     echo load $image to $node
-    docker save $image | bzip2 | pv | ssh -oStrictHostKeyChecking=no ec2-user@$node 'docker load'
```
reachjainrahul, line 113: if bzip2 is not required, can you update the doc and install_comon.sh to remove bzip2 dependency..

hack/terraform/eks/main.tf
```diff
@@ -100,7 +100,7 @@ module "eks" {
   version         = "~>17.24.0"
   cluster_name    = local.cluster_name
   subnets         = module.vpc.public_subnets
-  cluster_version = "1.23"
+  cluster_version = "1.25"
```
reachjainrahul, line 103: Could you pin 1.25.5 for EKS too

hack/eks
Anandkumar26: Updated it.

hack/terraform/eks/main.tf
Anandkumar26: Looks like EKS does not support minor versions. I did a quick try and it is failing. 
We can revisit this later.

```
│ Error: Your query returned no results. Please change your search criteria and try again.
│ 
│   with module.eks.data.aws_ami.eks_worker[0],
│   on .terraform/modules/eks/data.tf line 20, in data "aws_ami" "eks_worker":
│   20: data "aws_ami" "eks_worker" {
```

reachjainrahul (approved):
/LGTM

PR #146, opened 2023-03-07
fix vmp stuck in-progress when anp has no rules

shenmo3:
## Description
This PR fixes the bug where VMP is stuck in `in-progress` state if an ANP with no rules is applied.

## Changes
1. Rework the np status logic.


reachjainrahul:
nephe-test-e2e-azure

reachjainrahul (approved):
/LGTM

PR #150, opened 2023-03-08
add region field to vm crd

shenmo3:
## Description
Currently VM CRD doesn't have a field to indicate its cloud region, this PR adds such field.

## Changes
1. Add region field in VM CRD.
2. Re-generate relevant yaml files.
3. Update CR construct funcs and docs.

### DIFF
--- a/apis/crd/v1alpha1/virtualmachine_types.go
+++ b/apis/crd/v1alpha1/virtualmachine_types.go
@@ -86,6 +86,8 @@ type VirtualMachineStatus struct {
 	NetworkInterfaces []NetworkInterface `json:"networkInterfaces,omitempty"`
 	// State indicates current state of the VirtualMachine.
 	State VMState `json:"state,omitempty"`
+	// Region indicates the cloud region of the VirtualMachine.
+	Region string `json:"region,omitempty"`
 	// Agented specifies if VM runs in agented mode, default is false.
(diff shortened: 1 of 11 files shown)


apis/crd/v1alpha1/virtualmachine_types.go
```diff
@@ -86,6 +86,8 @@ type VirtualMachineStatus struct {
 	NetworkInterfaces []NetworkInterface `json:"networkInterfaces,omitempty"`
 	// State indicates current state of the VirtualMachine.
 	State VMState `json:"state,omitempty"`
+	// Region indicates the cloud region of the VirtualMachine.
```
reachjainrahul, line 89: Can you update the doc and also the tabular output of VM

reachjainrahul: Add +kubebuilder:printcolumn:name="Region",type=string,JSONPath=`.status.region` after Cloud-Provider so that VM output includes region and update getting started doc.

shenmo3: Done. Also `Agented` field is missing from doc get vm output, added that too.

reachjainrahul (approved):
/LGTM

reachjainrahul (approved):
/LGTM

PR #161, opened 2023-03-17
Remove deprecated actions in github workflow

reachjainrahul:
Use alexellis/upload-assets instead of actions/upload-release-asset, which is no longer maintained.

Remove deprecated set-output command.


Anandkumar26 (approved):
/LGTM

PR #164, opened 2023-03-20
Add checksum annotation on helm deployment yaml

reachjainrahul:
With checksum annotation we can handle pod restart if config map changes between helm charts


Anandkumar26 (approved):
/LGTM

PR #168, opened 2023-03-21
Add upgrade e2e test with helm

reachjainrahul:
## Description
Add upgrade e2e tests with helm version v0.5.

## Changes
Added network policy e2e upgrade tests. Base version is pinned to v0.5 and it will test upgrade with latest main branch. Tests also includes an option to test between release builds too.

### DIFF
--- a/ci/jenkins/scripts/test-aws.sh
+++ b/ci/jenkins/scripts/test-aws.sh
@@ -28,7 +28,8 @@ Setup and run integration tests on Kind cluster with AWS VMs.
         [--aws-service-user <ServiceUserName>]               AWS Service User Name.
         [--aws-service-user-role-arn <ServiceUserRoleARN>]   AWS Service User Role ARN.
         [--aws-region <Region>]                              AWS region where the setup will be deployed. Defaults to us-east-2.
-        [--owner <OwnerName>]                                Setup will be prefixed with owner name."
+        [--owner <OwnerName>]                                Setup will be prefixed with owner name.
+        [--upgrade]                                          Run upgrade test."
 
 function print_usage {
     echoerr "$_usage"
@@ -41,6 +42,7 @@ function print_help {
 # Defaults
 export TF_VAR_owner="ci"
 export AWS_DEFAULT_REGION="us-east-2"
+export UPGRADE=false 
 
 while [[ $# -gt 0 ]]
 do
@@ -70,6 +72,10 @@ case $key in
     --owner)
     export TF_VAR_owner="$2"
     shift 2
+    ;;
+    --upgrade)
+    export UPGRADE=true
+    shift 1
     ;;
     -h|--help)
     print_usage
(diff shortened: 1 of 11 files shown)


test/utils/networkpolicy_helper.go
```diff
@@ -0,0 +1,74 @@
...
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+
+	"antrea.io/nephe/pkg/labels"
+	k8stemplates "antrea.io/nephe/test/templates"
+)
+
+// ConfigANPApplyTo helper function to configure appliedTo in Antrea NetworkPolicy.
+func ConfigANPApplyTo(kind, instanceName, vpc, tagKey, tagVal string) *k8stemplates.EntitySelectorParameters {
```
shenmo3, line 27: Why splitting these two func into a separate file? Following this logic should we also split other funcs in np e2e tests like verifyXXX into this file also? This way upgrade test can also reuse some helper funcs. Personally I'm not a fan of current inline funcs, since they are a bit hard to track and IDEs doesn't highlight them. No need to be done in this PR though.

ci/jenkins/scripts/test-aws.sh
```diff
@@ -140,15 +146,22 @@ ci/kind/kind-setup.sh create kind
...
-    aws ec2 delete-key-pair  --key-name ${KEY_PAIR}  --region ${AWS_DEFAULT_REGION}
+    aws ec2 delete-key-pair  --key-name ${KEY_PAIR}  --region "${AWS_DEFAULT_REGION}"
 }
 trap cleanup EXIT
 
-mkdir -p $HOME/logs
-ci/bin/integration.test -ginkgo.v -ginkgo.timeout 90m -ginkgo.focus=".*test-aws.*" -kubeconfig=$HOME/.kube/config -cloud-provider=AWS -support-bundle-dir=$HOME/logs
+mkdir -p "$HOME"/logs
+
+if [ "$UPGRADE" = true ] ; then
```
shenmo3, line 161: Would it make more sense to make upgrades a separate focus, like `test-upgrade-aws/azure` instead of a separate suite?

test/upgrade/upgrade_suite_test.go
```diff
@@ -0,0 +1,152 @@
...
+	if !cloudVPC.IsConfigured() {
+		err := cloudVPC.Reapply(time.Second*600, false)
+		Expect(err).ToNot(HaveOccurred())
+		logf.Log.Info("VM VPCs created", "Provider", cloudProvider, "VPCID", cloudVPC.GetVPCID())
+	}
+
+})
+
+var _ = AfterSuite(func() {
+	if preserveSetup == true || kubeCtl == nil {
```
shenmo3, line 141: What's the difference between `preserveSetup` and `preserveSetupOnFail`?

ci/jenkins/scripts/test-aws.sh
reachjainrahul: We cannot be that.. Upgrade suite doesnt take nephe manifest. It start with 0.5 version and then upgrade to TOT. mixing with integration wasn't clean.

test/utils/networkpolicy_helper.go
reachjainrahul: This was used as is in both files, so I placed them in common. This is config, should be place. Verify I didnt put, because integration NP test have can more detailed verification step, then upgrade. Wanted to keep upgrade test simple. Note: This is not the final steps in upgrade. We will add more tests in upgrade tests.

test/upgrade/upgrade_suite_test.go
reachjainrahul: Think of preserveSetupOnFail as config. and preserveSetup as runtime variable. preserveSetup will be set and reset in each test suite. This will indicate to upgrade suite, whether to delete the VM cluster or keep it.

test/upgrade/upgrade_suite_test.go
```diff
@@ -0,0 +1,152 @@
...
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	antreav1alpha1 "antrea.io/antrea/pkg/apis/crd/v1alpha1"
+	antreav1alpha2 "antrea.io/antrea/pkg/apis/crd/v1alpha2"
+	cloudv1alpha1 "antrea.io/nephe/apis/crd/v1alpha1"
```
Anandkumar26, line 33: nit:
```suggestion
	crdv1alpha1 "antrea.io/nephe/apis/crd/v1alpha1"
```

test/utils/helpers.go
```diff
@@ -93,6 +93,16 @@ func StartOrWaitDeployment(k8sClient client.Client, name, namespace string, repl
 	return nil
 }
 
+// CheckIfConfigured checks if the deployment is configured.
+func CheckIfConfigured(k8sClient client.Client, name, namespace string) error {
```
Anandkumar26, line 97: Also, do we need a timeout here?
```suggestion
func CheckDeploment(k8sClient client.Client, name, namespace string) error {
```

reachjainrahul: Done.. Changed to CheckDeploymentConfigured. We just want to check if nephe deployment exist or not. No timeout is needed

test/upgrade/upgrade_suite_test.go
reachjainrahul: Done

shenmo3 (approved):

PR #178, opened 2023-03-28
fix vmp in progress on nephe controller restart

### LINKED ISSUE #177: vmp goes into in-progress state briefly after nephe controller restart with anp applied
**Describe the bug**
When the Nephe controller is restarted with an ANP applied, the vmp goes into the in-progress state briefly, even when there are no changes in the cloud.

**To Reproduce**
1. Import some VMs and apply an ANP.
2. Restart the Nephe controller.
3. Check the controller log.

**Expected**
The vmp should remain in the applied state, as there are no changes to the cloud.

**Actual behavior**
The controller log shows that the vmp briefly goes into the in-progress state, then returns to the applied state.

**Versions:**
v0.4.0

**Additional context**
N/A

shenmo3:
## Description
When the Nephe controller is restarted with ANPs applied, the vmp briefly enters the 'in-progress' state for less than a second before returning to the 'applied' state. This behavior can be misleading, as the ANP is still applied on the cloud and Nephe is simply resyncing the state. This PR fixes this issue by postponing the vmp updates until the initial cloud sync is completed.

## Changes
1. Do not update the vmp status until initial cloud sync is completed.

Closes #177

### DIFF
--- a/pkg/controllers/cloud/networkpolicy_cloudresource.go
+++ b/pkg/controllers/cloud/networkpolicy_cloudresource.go
@@ -135,6 +135,7 @@ func (r *NetworkPolicyReconciler) newCloudResourceNPTracker(rsc *securitygroup.C
 		prevAppliedToSGs: make(map[string]*appliedToSecurityGroup),
 		cloudResource:    *rsc,
 	}
+	tracker.unmarkDirty()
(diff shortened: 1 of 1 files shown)


reachjainrahul:
Fixed.

Issue #183, opened 2023-03-29
With Invalid credentials, CPA CR add is retried indefinitely 


reachjainrahul:
**Describe the bug**

If account credentials are wrong, CPA controller tries to handle the add indefinitely

**To Reproduce**

Add CPA CR with invalid credentials

**Expected**

Store invalid credentials state in CPA object and bail out

**Versions:**

v0.4.0

**Additional context**

```
2023-03-29T03:42:44.317Z        INFO    aws-plugin      aws/aws_ec2.go:377      cache snapshot nil      {"service": "EC2", "account": "demo-ns/aws-account"}
2023-03-29T03:42:44.317Z        DEBUG   aws-plugin      internal/services_common.go:124 cached vpcs     {"service": "EC2", "account": "demo-ns/aws-account", "vpc objects": 0}
2023-03-29T03:42:44.317Z        INFO    poller.AccountPoller    wait/wait.go:155        Discovered compute resources statistics {"account": "demo-ns/aws-account", "vpcs": 0, "virtual-machines": 0}
2023-03-29T03:42:44.318Z        INFO    controllers.CloudProviderAccount        cloud/cloudprovideraccount_controller.go:77     Received request        {"account": "demo-ns/aws-account", "operation": "create/update"}
2023-03-29T03:42:44.323Z        INFO    aws-plugin      internal/accounts_common.go:127 credentials not changed.        {"account": "demo-ns/aws-account"}
2023-03-29T03:42:44.323Z        INFO    poller.AccountPoller    cloud/cloudprovideraccount_controller.go:148    Poller exists   {"account": "demo-ns/aws-account"}
2023-03-29T03:42:44.323Z        INFO    poller.AccountPoller    cloud/cloudprovideraccount_controller.go:153    Restarting account poller       {"account": "demo-ns/aws-account"}
2023-03-29T03:42:44.340Z        ERROR   aws-plugin      internal/services_common.go:92  failed to fetch cloud resources {"account": "demo-ns/aws-account", "error": "error describing ec2 vpcs: \"AccessDenied: User: arn:aws:sts::XXX:assumed-role/nephe-eks-node-ec2-role/i-092ffbd7e29aac217 is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::XXX:role/cloud-controller-ci-role\\n\\tstatus code: 403, request id: 27b5c6a7-1c47-4161-a5be-023dcd583b2e\""}
[...]

PR #186, opened 2023-03-30
Capture CPA reconciler errors in status field

Anandkumar26:
- Do not return error back to the reconciler, if the error can not be fixed upn retry. Instead update the status field with the error message. Also clean up error messages.
    
- When a CPA is updated/created, the generation is incremented. when the CPA status is updated, the generation is unchanged. Allow CPA controller to receive CPA updates for all events except a change in status.
    
Fix: https://github.com/antrea-io/nephe/issues/183


pkg/controllers/cloud/cloudprovideraccount_controller.go
```diff
@@ -84,7 +86,9 @@ func (r *CloudProviderAccountReconciler) Reconcile(ctx context.Context, req ctrl
 
 func (r *CloudProviderAccountReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	r.accountProviderType = make(map[types.NamespacedName]common.ProviderType)
-	if err := ctrl.NewControllerManagedBy(mgr).For(&crdv1alpha1.CloudProviderAccount{}).Complete(r); err != nil {
+	if err := ctrl.NewControllerManagedBy(mgr).
+		For(&crdv1alpha1.CloudProviderAccount{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
```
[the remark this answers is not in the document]
Anandkumar26: Done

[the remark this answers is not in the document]
reachjainrahul: be specific.. we are using GenerationChangedPredicate to allow CPA controller to receive CPA updates for all events except change in status.

reachjainrahul (approved):
/LGTM

pkg/controllers/cloud/cloudprovideraccount_controller.go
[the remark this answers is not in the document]
Anandkumar26: Done

pkg/accountmanager/manager.go
```diff
@@ -241,7 +235,7 @@ func (a *AccountManager) addAccountPoller(cloudInterface common.CloudInterface,
 func (a *AccountManager) removeAccountPoller(namespacedName *types.NamespacedName) error {
 	accPoller, exists := a.getAccountPoller(namespacedName)
 	if !exists {
-		return nil
+		return fmt.Errorf(fmt.Sprintf("%v %v", errorMsgAccountPollerNotFound, namespacedName))
```
reachjainrahul, line 238: why this error message. if it doesnt exist, let it be, as its remove case

pkg/controllers/cloudprovideraccount/controller.go
```diff
@@ -131,9 +137,16 @@ func (r *CloudProviderAccountReconciler) processCreateOrUpdate(namespacedName *t
 	r.Log.Info("Received request", "account", namespacedName, "operation", "create/update")
 	accountCloudType, err := util.GetAccountProviderType(account)
 	if err != nil {
-		return fmt.Errorf("%s, err: %v", accountmanager.ErrorMsgAccountAddFail, err)
+		return fmt.Errorf("%s: %v", accountmanager.ErrorMsgAddOrUpdateAccount, err)
+	}
+	if err = r.AccManager.AddAccount(namespacedName, accountCloudType, account); err != nil {
+		account.Status.Error = err.Error()
```
reachjainrahul, line 143: this will retry reconciler right?

pkg/controllers/cloudprovideraccount/controller.go
```diff
@@ -183,36 +196,48 @@ func (r *CloudProviderAccountReconciler) getCpaBySecret(s types.NamespacedName)
...
+			case watch.Deleted:
+				r.mutex.Lock()
+				logWithSecret.Info("Received request", "Secret", namespacedName, "operation", "delete")
+				cpaItems, err := r.getCpaBySecret(namespacedName)
+				if err != nil {
+					logWithSecret.Error(err, "error getting CPA by Secret", "account", namespacedName)
+				}
+				for _, cpa := range cpaItems {
+					accountNamespacedName := types.NamespacedName{Namespace: cpa.Namespace, Name: cpa.Name}
+					logWithSecret.Info("Done processing delete", "account", accountNamespacedName)
```
reachjainrahul, line 238: info log is not clear.. you have to say, secret object is getting removed, however there is CPA referencing to it. Please delete it.. Also, not sure, if you need locks here.

pkg/controllers/cloudprovideraccount/controller.go
Anandkumar26: No. I want to capture any errors as part of AddAccount in CPA status field, so that user can check the CPA status field.

In L145, if there is any error as part of updating status field, then we will lose the error context. So we will be return error and let the reconciler retry.

pkg/controllers/cloudprovideraccount/controller.go
Anandkumar26: I kept the lock here to make sure when watcher would be doing one operation at a time.  On a second thought, lock is not required, since everything is being processing inline. I will remove the lock.

pkg/accountmanager/manager.go
Anandkumar26: On the CPA remove account, we will be ignoring any error returned as part of removeAccountPoller.

Going by the function name removeAccountPoller - expect to remove the poller.  If for some reason, we are unable then we can return the error to the caller so that he can chose to ignore/handle it.

pkg/controllers/cloudprovideraccount/controller.go
Anandkumar26: is this better?
```suggestion
										logWithSecret.Info("Done processing Secret delete, update/delete stale account", "account", accountNamespacedName)

```

reachjainrahul (approved):
/LGTM

PR #189, opened 2023-04-03
Add pending queue for network Policy

reachjainrahul:
Pending queue will hold network policy event if rules are getting updated in cloud. Without this, new networkPolicy event would operate on stale cloudRule Indexer as previous rule update in ongoing in cloud.

### DIFF
--- a/pkg/controllers/cloud/networkpolicy.go
+++ b/pkg/controllers/cloud/networkpolicy.go
@@ -109,6 +109,10 @@ const (
 	uniqueGroupNameMemberPrefix = "mm_"
 )
 
+const (
+	pendingQBufferLimit = 50
+)
+
 func getGroupUniqueName(name string, memberOnly bool) string {
 	if memberOnly {
 		return uniqueGroupNameMemberPrefix + name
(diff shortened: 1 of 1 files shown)


pkg/controllers/cloud/networkpolicy.go
```diff
@@ -292,6 +296,8 @@ type securityGroupImpl struct {
 	retryOp *securityGroupOperation
 	// true if retry operation is ongoing.
 	retryInProgress bool
+	// true if cloud operation is ongoing.
+	cloudOpInProgress bool
```
Anandkumar26, line 300: This will be extended to track any on-going cloud operation right. (Even for addressGroup)

pkg/controllers/cloud/networkpolicy.go
```diff
@@ -109,6 +109,10 @@ const (
 	uniqueGroupNameMemberPrefix = "mm_"
 )
 
+const (
```
Anandkumar26, line 112: nit: may be add this in other const block

reachjainrahul: type is different, dont want to overload other block

Anandkumar26 (approved):
/LGTM

pkg/controllers/cloud/networkpolicy.go
reachjainrahul: no, just for AppliedToGroup.. Later we might add types for different operation based on new VMP design

PR #190, opened 2023-04-03
Minimize the use of resourceGraph query for interface data

reachjainrahul:
It was been observed resource graph query doesnt give latest data of the network interface. For example, if ASG is getting attached on the interface and immediately in the next call, we try to update the interface with some other ASG, that interface may or may not have the previous ASG, leading to misleading update.

Here we use ResourceGraph query only to give interface to vnet mapping which is used for filtering in getInterfaces call.

### DIFF
--- a/pkg/cloud-provider/cloudapi/azure/azure_security.go
+++ b/pkg/cloud-provider/cloudapi/azure/azure_security.go
@@ -38,7 +38,13 @@ var (
 	mutex sync.Mutex
 )
 
-func (computeCfg *computeServiceConfig) getNetworkInterfacesOfVnet(vnetIDSet map[string]struct{}) ([]*networkInterfaceTable, error) {
+type networkInterfaceInternal struct {
+	armnetwork.Interface
+	vnetID string
+}
+
+// getNetworkInterfacesOfVnet fetch network interfaces for a set of VNET-IDs.
+func (computeCfg *computeServiceConfig) getNetworkInterfacesOfVnet(vnetIDSet map[string]struct{}) ([]*networkInterfaceInternal, error) {
 	location := computeCfg.credentials.region
 	subscriptionID := computeCfg.credentials.SubscriptionID
 	tenentID := computeCfg.credentials.TenantID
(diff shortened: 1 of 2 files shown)


pkg/cloud-provider/cloudapi/azure/azure_security.go
```diff
@@ -83,51 +113,58 @@ func (computeCfg *computeServiceConfig) processAppliedToMembership(appliedToGrou
...
+					_, _, isAT := securitygroup.IsNepheControllerCreatedSG(asgName)
+					if isAT {
+						if strings.Compare(nsgNameLowercase, perVnetNsgNameLowercase) == 0 &&
+							strings.Compare(cloudSgNameLowercase, asgName) == 0 {
+							isNsgAttached = true
+							break
+						}
 					}
 				}
+				// Handling just first available ip-configuration; with an assumption that it will be primary IP always.
```
Anandkumar26, line 154: Is this a TODO? or currently does the first index always returns primary IP config?

pkg/cloud-provider/cloudapi/azure/azure_security.go
```diff
@@ -83,51 +113,58 @@ func (computeCfg *computeServiceConfig) processAppliedToMembership(appliedToGrou
...
 			}
 		}
-		_, isNicAttachedToMemberVM := memberVirtualMachines[strings.ToLower(*networkInterface.VirtualMachineID)]
+
+		_, isNicAttachedToMemberVM := memberVirtualMachines[strings.ToLower(*networkInterface.Properties.VirtualMachine.ID)]
 		_, isNicMemberNetworkInterface := memberNetworkInterfaces[strings.ToLower(*networkInterface.ID)]
 		if isNsgAttached {
 			if !isNicAttachedToMemberVM && !isNicMemberNetworkInterface {
-				nwIntfIDSetNsgToDetach[nwIntfIDLowerCase] = struct{}{}
+				nwIntfIDSetNsgToDetach[nwIntfIDLowerCase] = &networkInterface.Interface
```
Anandkumar26, line 163: Here we are assigning the address of a local variable, the memory can get overiden when reading 2nd or 3rd interfaces.. 

This has to be changed like this.
for i := range networkInterfaces {
&networkInterfaces[i]

pkg/cloud-provider/cloudapi/azure/azure_security.go
```diff
@@ -192,85 +221,85 @@ func (computeCfg *computeServiceConfig) processAddressGroupMembership(addressGro
...
 				continue
 			}
-			_, isNepheControllerCreatedAG, _ := securitygroup.IsNepheControllerCreatedSG(asgNameLowercase)
-			if !isNepheControllerCreatedAG {
-				continue
-			}
-			if strings.Compare(asgNameLowercase, cloudAsgNameLowercase) == 0 {
-				isAsgAttached = true
+			for _, asg := range ipconfig.Properties.ApplicationSecurityGroups {
+
```
Anandkumar26: nit: remove line.

pkg/cloud-provider/cloudapi/azure/azure_security.go
reachjainrahul: yes.. I verified

pkg/cloud-provider/cloudapi/azure/azure_security.go
reachjainrahul: not really.. &networkInterface.Interface.. networkInterface is of type 
type networkInterfaceInternal struct {
	armnetwork.Interface
	vnetID string
} and address of armnetwork.Interface is passed

pkg/cloud-provider/cloudapi/azure/azure_security.go
reachjainrahul: removed

pkg/cloud-provider/cloudapi/azure/azure_security.go
Anandkumar26: If it was just &networkInterface, then that could have caused issue, but here it is address on struct member, so it would be fine.

PR #192, opened 2023-04-03
Update codecov.yaml

reachjainrahul:
moving the patch part in coverage. It gives misleading info. We will only check if the overall coverage drops by 2 percent


Anandkumar26 (approved):
/LGTM

PR #193, opened 2023-04-03
Reduce AppliedTo state handling

reachjainrahul:
AppliedToGroup delete doesnt depend on NetworkPolicy delete. Hence removing the check in deletion and reduce unneccessary notify trigger from networkPolicy delete/update.


Anandkumar26 (approved):
/LGTM

PR #198, opened 2023-04-03
Fix azure terraform rule naming issue

shenmo3:
## Description
Azure recently changes their nsg rule naming restriction, not allowing `*` in names. In our Azure agented test, the VMs will allow `*` inbounds and the Azure module will use the `*` in the rule name, causing error. Until the fix is released in the module, I propose a temporary fix which adds the allow `*` rule manually.

## Changes
1. Disable allow `*` rule in agented test in Azure module.
2. Add allow `*` rule resource after module is applied.


reachjainrahul (approved):
/LGTM

PR #210, opened 2023-04-13
Update dependent CPAs for Secret add/modify/delete operation

Anandkumar26:
- Add: Updates account only if its error state with error as Secret does not exist.
- Update/Delete: Updates account to recompute account state, as Secret is changed
- Cleanup account state when account add/update fails.

### DIFF
--- a/pkg/accountmanager/poller.go
+++ b/pkg/accountmanager/poller.go
@@ -161,10 +161,7 @@ func (p *accountPoller) doAccountPolling() {
 			p.namespacedName.String())
 		return
 	}
-
-	if err := p.inventory.BuildVpcCache(vpcMap, p.namespacedName); err != nil {
-		p.log.Error(err, "failed to build vpc cache", "account", p.namespacedName.String())
-	}
+	_ = p.inventory.BuildVpcCache(vpcMap, p.namespacedName)
 
 	// Perform VM Operations only when CES is added.
 	vmCount := 0
@@ -184,24 +181,23 @@ func (p *accountPoller) doAccountPolling() {
 // updateAccountStatus updates status of a CPA object when it's changed.
 func (p *accountPoller) updateAccountStatus(cloudInterface common.CloudInterface) {
 	account := &crdv1alpha1.CloudProviderAccount{}
-	e := p.Get(context.TODO(), *p.namespacedName, account)
-	if e != nil {
+	err := p.Get(context.TODO(), *p.namespacedName, account)
+	if err != nil {
 		return
 	}
 
 	discoveredStatus := crdv1alpha1.CloudProviderAccountStatus{}
-	status, e := cloudInterface.GetAccountStatus(p.namespacedName)
-	if e != nil {
-		discoveredStatus.Error = fmt.Sprintf("failed to get status, err %v", e)
+	status, err := cloudInterface.GetAccountStatus(p.namespacedName)
+	if err != nil {
+		discoveredStatus.Error = fmt.Sprintf("failed to get status, err %v", err)
 	} else if status != nil {
 		discoveredStatus = *status
 	}
 
 	if account.Status != discoveredStatus {
 		account.Status.Error = discoveredStatus.Error
-		e = p.Client.Status().Update(context.TODO(), account)
-		if e != nil {
-			p.log.Error(e, "failed to update account status", "account", p.namespacedName)
+		if err = p.Client.Status().Update(context.TODO(), account); err != nil {
+			p.log.Error(err, "failed to update account status", "account", p.namespacedName)
 		}
 	}
 }
--- a/pkg/controllers/cloudprovideraccount/controller.go
+++ b/pkg/controllers/cloudprovideraccount/controller.go
@@ -17,6 +17,7 @@ package cloudprovideraccount
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -137,7 +138,7 @@ func (r *CloudProviderAccountReconciler) processCreateOrUpdate(namespacedName *t
 	r.Log.Info("Received request", "account", namespacedName, "operation", "create/update")
 	accountCloudType, err := util.GetAccountProviderType(account)
 	if err != nil {
-		return fmt.Errorf("%s: %v", accountmanager.ErrorMsgAddOrUpdateAccount, err)
+		return fmt.Errorf("failed to add or update account: %v", err)
 	}
 	if err = r.AccManager.AddAccount(namespacedName, accountCloudType, account); err != nil {
 		account.Status.Error = err.Error()
@@ -145,6 +146,7 @@ func (r *CloudProviderAccountReconciler) processCreateOrUpdate(namespacedName *t
 		if err = r.Client.Status().Update(context.TODO(), account); err != nil {
 			return fmt.Errorf("failed to update account status, account %v err %v", namespacedName, err)
 		}
+		_ = r.AccManager.RemoveAccount(namespacedName)
 	}
 	return nil
 }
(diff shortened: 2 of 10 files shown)


pkg/controllers/cloudprovideraccount/controller.go
```diff
@@ -175,70 +177,65 @@ func (r *CloudProviderAccountReconciler) setSyncStatusAndSecretWatcher() {
...
+
+	for _, cpa := range cpaItems {
+		accountNamespacedName := &types.NamespacedName{Namespace: cpa.Namespace, Name: cpa.Name}
+		// For add operation, update CPA only when status contains an error ErrorMsgSecretDoesNotExist.
+		if eventType == watch.Added {
+			if !strings.Contains(cpa.Status.Error, util.ErrorMsgSecretDoesNotExist) {
+				continue
+			}
+		}
+		if err := r.processCreateOrUpdate(accountNamespacedName, cpa); err != nil {
```
archanapholla, line 222: processCreateOrUpdate function can be called from reconciler too. Let's say, before this function acquires the lock, reconciler calls processCreateOrUpdate for the same CPA(CPA attached to the secret in this function). 
When the execution is complete and this function acquires the lock, this might be operating on stale secret event.

Do you see any issue in this scenario?

pkg/controllers/cloudprovideraccount/controller.go
```diff
@@ -175,70 +177,65 @@ func (r *CloudProviderAccountReconciler) setSyncStatusAndSecretWatcher() {
...
-				cpaItems = append(cpaItems, cpa)
+				cpaItems = append(cpaItems, &cpa)
 			}
 		}
 	}
 	return cpaItems, nil
 }
 
-// watchSecret watch the Secret objects.
+// updateDependentCpaBySecretNamespace updates all dependent accounts for a given Secret Namespace.
```
archanapholla, line 203: Should the comment state "for a given Secret" instead of Secret namespace

pkg/controllers/cloudprovideraccount/controller.go
Anandkumar26: At any point of time only one of them will acquire the lock and update the account's internal state.

You are talking about a scenario where both events occur at the same time.
- update of CPA object, which triggers reconciler. 
- update on the Secret object.

In this particular scenario,  processCreateOrUpdate will be called twice(once via Secret and once via reconciler), as a result we would restart poller and perform polling twice(which cannot be avoided).

I don't see any issue here.

pkg/controllers/cloudprovideraccount/controller.go
Anandkumar26: No. Secret watcher watches for Secrets only from POD_NAMESPACE. This function will be called only for the Secret namespace.

pkg/accountmanager/poller.go
```diff
@@ -199,8 +196,7 @@ func (p *accountPoller) updateAccountStatus(cloudInterface common.CloudInterface
 
 	if account.Status != discoveredStatus {
 		account.Status.Error = discoveredStatus.Error
-		e = p.Client.Status().Update(context.TODO(), account)
-		if e != nil {
+		if e = p.Client.Status().Update(context.TODO(), account); e != nil {
```
reachjainrahul, line 199: rename e to err?

pkg/controllers/cloudprovideraccount/controller.go
```diff
@@ -137,14 +138,15 @@ func (r *CloudProviderAccountReconciler) processCreateOrUpdate(namespacedName *t
...
-		return fmt.Errorf("%s: %v", accountmanager.ErrorMsgAddOrUpdateAccount, err)
+		return fmt.Errorf("failed to add or update account: %v", err)
 	}
 	if err = r.AccManager.AddAccount(namespacedName, accountCloudType, account); err != nil {
 		account.Status.Error = err.Error()
 		r.Log.Info("Setting account status", "account", namespacedName, "err", err)
 		if err = r.Client.Status().Update(context.TODO(), account); err != nil {
 			return fmt.Errorf("failed to update account status, account %v err %v", namespacedName, err)
 		}
+		_ = r.AccManager.RemoveAccount(namespacedName)
```
reachjainrahul, line 149: do we really need to call removal? the AddAccount will be retried right?

pkg/controllers/cloudprovideraccount/controller.go
```diff
@@ -175,70 +177,65 @@ func (r *CloudProviderAccountReconciler) setSyncStatusAndSecretWatcher() {
...
+	defer r.mutex.Unlock()
+
+	cpaItems, err := r.getCpaBySecret(secretNamespacedName)
+	if err != nil {
+		r.Log.WithName("Secret").Error(err, "error getting CPA by Secret", "Secret", secretNamespacedName)
+	}
+
+	for _, cpa := range cpaItems {
+		accountNamespacedName := &types.NamespacedName{Namespace: cpa.Namespace, Name: cpa.Name}
+		// For add operation, update CPA only when status contains an error ErrorMsgSecretDoesNotExist.
```
reachjainrahul, line 216: there is no handling for delete, how are you setting ErrorMsgSecretDoesNotExist in CPA?

pkg/controllers/cloudprovideraccount/controller.go
```diff
@@ -175,70 +177,65 @@ func (r *CloudProviderAccountReconciler) setSyncStatusAndSecretWatcher() {
...
-				for _, cpa := range cpaItems {
-					accountNamespacedName := types.NamespacedName{Namespace: cpa.Namespace, Name: cpa.Name}
-					logWithSecret.Info("Done processing Secret delete, update/delete stale account", "account", accountNamespacedName)
-				}
-			case watch.Error:
-				r.resetSecretWatcher()
-			}
+			namespacedName := &types.NamespacedName{Namespace: secret.Namespace, Name: secret.Name}
+			r.Log.WithName("Secret").Info("Received request", "Secret", namespacedName, "operation", event.Type)
+			r.updateDependentCpaBySecretNamespace(namespacedName, event.Type)
```
reachjainrahul, line 238: call it processSecretUpdateEvent..

pkg/controllers/cloudprovideraccount/controller.go
```diff
@@ -175,70 +177,65 @@ func (r *CloudProviderAccountReconciler) setSyncStatusAndSecretWatcher() {
...
+
+	cpaItems, err := r.getCpaBySecret(secretNamespacedName)
+	if err != nil {
+		r.Log.WithName("Secret").Error(err, "error getting CPA by Secret", "Secret", secretNamespacedName)
+	}
+
+	for _, cpa := range cpaItems {
+		accountNamespacedName := &types.NamespacedName{Namespace: cpa.Namespace, Name: cpa.Name}
+		// For add operation, update CPA only when status contains an error ErrorMsgSecretDoesNotExist.
+		if eventType == watch.Added {
```
reachjainrahul, line 217: you dont need to handle watch.Delete or modify?

pkg/controllers/cloudprovideraccount/controller.go
```diff
@@ -175,70 +177,65 @@ func (r *CloudProviderAccountReconciler) setSyncStatusAndSecretWatcher() {
...
+			if !strings.Contains(cpa.Status.Error, util.ErrorMsgSecretDoesNotExist) {
+				continue
```
reachjainrahul, line 219: Add a log here

pkg/controllers/cloudprovideraccount/controller.go
Anandkumar26: Done

pkg/controllers/cloudprovideraccount/controller.go
Anandkumar26: Done

pkg/accountmanager/poller.go
Anandkumar26: Done.

pkg/controllers/cloudprovideraccount/controller.go
Anandkumar26: Yes. We don't want to reconciler to keep trying whenever there is an error returned by CPA. 
If for any reason, we are not able to process CPA request, we will set the error status on the CPA object and clean up our internal state.

reachjainrahul (approved):
/LGTM

PR #211, opened 2023-04-13
Added sort function for vpc and vm output

reachjainrahul:
Also removed omitempty from managed field in Vpc Object to correctly indicate its status in Vpc get call.


Anandkumar26 (approved):
/LGTM

PR #217, opened 2023-04-26
Remove owner reference from ExternalNode CR

Anandkumar26:
VM is a runtime object, which gets re-created upon restart. The owner reference on  the ExternlaNode CR will not exist, so it is deleted by kubernetes run time.

Without owner references on ExternalNode CR, any event on the VM, will be fed to the convert module and convert will accordingly perform action if needed.


archanapholla (approved):
/LGTM

PR #220, opened 2023-05-05
preserve azure rule priority on rule update

### LINKED ISSUE #206: Azure Nephe rule priority change on networkpolicy modify
**Describe the bug**
When ANP rule changes for Azure VMs, existing rules' priority might change even there are no change related to those specific rules.

**To Reproduce**
1. Import some Azure VMs.
2. Apply an ANP.
3. Add additional rules to that ANP and re-apply it

**Expected**
When a new rule is added, previous rule's priority should not change and the new rule will be added after previous rule priorities at a higher number.

**Actual behavior**
Sometimes, the new rule will take old rule's priority number, and the old rule's priority will change to a new number.

**Versions:**
v0.4.0

**Additional context**
Multiple attempts might be needed, the behavior is not consistent.

This is also true for rule deletion when there are multiple rules.
e.g. two rules at priority 2000 and 2001, the 2000 rule is removed in ANP, Nephe will remove the 2000 rule and change 2001 rule's priority to 2000.

shenmo3:
## Description
Currently when rules are updated on Azure, the priorities of rules are recomputed instead of preserving the previous priority. This PR resolves this problem by checking if rule already exists on Azure rule update and skips priority computation if it already exists.

## Changes
1. Implement function to compare Azure security rules based on fields Nephe uses.
2. Change update rule priority logic to skip existing rules and skip used priorities when computing new rules. New rules will take the first available priority starting from Nephe priority range 2000.
3. Change rule update logic to use addRules and rmRules to perform delta update on existing rules. addRules will be passed into rule priority update logic to compute priority.
4. Remove allRules from `UpdateSecurityGroupRules` func parameter as Azure no longer uses it. Now AWS and Azure are consistent.
5. Update corresponding unit tests.

## Closes
#206

### DIFF
--- a/pkg/cloudprovider/cloudapi/azure/azure_nsg_rules.go
+++ b/pkg/cloudprovider/cloudapi/azure/azure_nsg_rules.go
@@ -17,6 +17,7 @@ package azure
 import (
 	"fmt"
 	"net"
+	"reflect"
 	"strconv"
 	"strings"
 
@@ -67,29 +68,33 @@ func isAzureRuleAttachedToAtSg(rule *armnetwork.SecurityRule, asg string) bool {
 	return false
 }
 
-// updateSecurityRuleNameAndPriority updates rule name and priority from existing
-// and new security rules and returns all the security rules.
-func updateSecurityRuleNameAndPriority(existingRules []armnetwork.SecurityRule,
-	newRules []armnetwork.SecurityRule) []armnetwork.SecurityRule {
-	var rules []armnetwork.SecurityRule
-	defaultAndUserCreatedRules := make([]armnetwork.SecurityRule, 0)
+// getUnusedPriority finds and returns the first unused priority starting from startPriority.
+func getUnusedPriority(existingRulePriority map[int32]struct{}, startPriority int32) int32 {
+	_, ok := existingRulePriority[startPriority]
+	for ok {
+		startPriority++
+		_, ok = existingRulePriority[startPriority]
+	}
+	return startPriority
+}
 
+// updateSecurityRuleNameAndPriority updates rule name and priority for new security rules based on existing security rules
+// and returns them combined.
+func updateSecurityRuleNameAndPriority(existingRules []*armnetwork.SecurityRule,
+	newRules []*armnetwork.SecurityRule) []*armnetwork.SecurityRule {
+	var rules []*armnetwork.SecurityRule
+	existingRulePriority := make(map[int32]struct{})
 	rulePriority := int32(ruleStartPriority)
+
 	for _, rule := range existingRules {
 		if rule.Properties == nil {
 			continue
 		}
-		// skip priority update for user created custom rules outside Nephe priority range.
-		if *rule.Properties.Priority < ruleStartPriority {
-			defaultAndUserCreatedRules = append(defaultAndUserCreatedRules, rule)
-			continue
+		// record priority for existing rules in Nephe priority range.
+		if *rule.Properties.Priority >= ruleStartPriority {
+			existingRulePriority[*rule.Properties.Priority] = struct{}{}
 		}
-		ruleName := fmt.Sprintf("%v-%v", rulePriority, *rule.Properties.Direction)
-		rule.Name = &ruleName
-		rule.Properties.Priority = to.Int32Ptr(rulePriority)
-
 		rules = append(rules, rule)
-		rulePriority++
 	}
 
 	for _, rule := range newRules {
@@ -97,19 +102,20 @@ func updateSecurityRuleNameAndPriority(existingRules []armnetwork.SecurityRule,
 			continue
 		}
 		// skip priority update for default 4096 deny rule. newRules consists of only Nephe rules.
-		if *rule.Properties.Priority == vnetToVnetDenyRulePriority {
-			defaultAndUserCreatedRules = append(defaultAndUserCreatedRules, rule)
+		if rule.Properties.Priority != nil && *rule.Properties.Priority == vnetToVnetDenyRulePriority {
+			rules = append(rules, rule)
 			continue
 		}
+
+		// update priority for new rules.
+		rulePriority = getUnusedPriority(existingRulePriority, rulePriority)
+		rule.Properties.Priority = to.Int32Ptr(rulePriority)
 		ruleName := fmt.Sprintf("%v-%v", rulePriority, *rule.Properties.Direction)
 		rule.Name = &ruleName
-		rule.Properties.Priority = to.Int32Ptr(rulePriority)
 
 		rules = append(rules, rule)
 		rulePriority++
 	}
-
-	rules = append(rules, defaultAndUserCreatedRules...)
 
 	return rules
 }
@@ -117,15 +123,16 @@ func updateSecurityRuleNameAndPriority(existingRules []armnetwork.SecurityRule,
 // convertIngressToNsgSecurityRules converts ingress rules from securitygroup.CloudRule to azure rules.
 func convertIngressToNsgSecurityRules(appliedToGroupID *securitygroup.CloudResourceID, rules []*securitygroup.CloudRule,
 	agAsgMapByNepheControllerName map[string]armnetwork.ApplicationSecurityGroup,
-	atAsgMapByNepheControllerName map[string]armnetwork.ApplicationSecurityGroup) ([]armnetwork.SecurityRule, error) {
-	var securityRules []armnetwork.SecurityRule
+	atAsgMapByNepheControllerName map[string]armnetwork.ApplicationSecurityGroup) ([]*armnetwork.SecurityRule, error) {
+	var securityRules []*armnetwork.SecurityRule
 
-	dstAsgObj, found := atAsgMapByNepheControllerName[strings.ToLower(appliedToGroupID.Name)]
+	asg, found := atAsgMapByNepheControllerName[strings.ToLower(appliedToGroupID.Name)]
 	if !found {
-		return []armnetwork.SecurityRule{}, fmt.Errorf("asg not found for applied to SG %v", appliedToGroupID.Name)
+		return []*armnetwork.SecurityRule{}, fmt.Errorf("asg not found for applied to SG %v", appliedToGroupID.Name)
 	}
+	// only resource id is needed. other fields are ignored to match Azure returned format so rules can be compared.
+	dstAsgObj := armnetwork.ApplicationSecurityGroup{ID: asg.ID}
 
-	rulePriority := int32(ruleStartPriority)
 	for _, obj := range rules {
 		rule := obj.Rule.(*securitygroup.IngressRule)
 		if rule == nil {
@@ -133,11 +140,11 @@ func convertIngressToNsgSecurityRules(appliedToGroupID *securitygroup.CloudResou
 		}
 		description, err := securitygroup.GenerateCloudDescription(obj.NpNamespacedName)
 		if err != nil {
-			return []armnetwork.SecurityRule{}, fmt.Errorf("unable to generate rule description, err: %v", err)
+			return []*armnetwork.SecurityRule{}, fmt.Errorf("unable to generate rule description, err: %v", err)
 		}
 		protoName, err := convertToAzureProtocolName(rule.Protocol)
 		if err != nil {
-			return []armnetwork.SecurityRule{}, err
+			return []*armnetwork.SecurityRule{}, err
 		}
 
 		srcPort := convertToAzurePortRange(rule.FromPort)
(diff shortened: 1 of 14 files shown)


pkg/cloudprovider/cloudapi/azure/azure_nsg_rules.go
```diff
@@ -51,63 +52,65 @@ func getDefaultDenyRuleName() string {
...
+	var securityRules []*armnetwork.SecurityRule
 
-	dstAsgObj, found := atAsgMapByNepheControllerName[strings.ToLower(appliedToGroupID.Name)]
+	asg, found := atAsgMapByNepheControllerName[strings.ToLower(appliedToGroupID.Name)]
 	if !found {
-		return []armnetwork.SecurityRule{}, fmt.Errorf("asg not found for applied to SG %v", appliedToGroupID.Name)
+		return []*armnetwork.SecurityRule{}, fmt.Errorf("asg not found for applied to SG %v", appliedToGroupID.Name)
 	}
+	// only resource id is needed. other fields are ignored to match Azure returned format so rules can be compared.
+	dstAsgObj := armnetwork.ApplicationSecurityGroup{ID: asg.ID}
```
Anandkumar26, line 113: Does this change help in any way?What do we gain by using only the ID field.

pkg/cloudprovider/cloudapi/azure/azure_security.go
```diff
@@ -320,53 +337,39 @@ func (computeCfg *computeServiceConfig) buildEffectiveNSGSecurityRulesToApply(ap
...
 			_, _, isNepheControllerCreatedRule := securitygroup.IsNepheControllerCreatedSG(desc.AppliedToGroup)
 			if isNepheControllerCreatedRule && strings.Compare(ruleAddrGroupName, appliedToGroupNepheControllerName) == 0 {
-				continue
+				// if the rule no longer exists in newRules then skip its update.
+				// if the rule exists then remove it from newRules and add to current rules.
+				newRules := newIngressSecurityRules
+				if *rule.Properties.Direction == armnetwork.SecurityRuleDirectionOutbound {
+					newRules = newEgressSecurityRules
+				}
+				idx, found := findSecurityRule(rule, newRules)
```
Anandkumar26, line 353: Here you are checking if the newRule and currentRule realized in cloud are same. 
If they are same, then you are going to skip updating the rule right?

pkg/cloudprovider/cloudapi/azure/azure_nsg_rules.go
```diff
@@ -409,6 +414,62 @@ func buildPeerSecurityRule(rulePriority *int32, protoName armnetwork.SecurityRul
 	return securityRule
 }
 
+// findSecurityRule finds the security rule in the given slice and return the index with boolean indicating found or not.
+func findSecurityRule(rule *armnetwork.SecurityRule, ruleList []*armnetwork.SecurityRule) (int, bool) {
```
Anandkumar26, line 418: This function not only finds the rule, it also checks if the rule is changed.
Please change the function name accordingly.

shenmo3: You mean check the rule list is changed? I don't think so, changing the rule in list to nil is just an optimization, the goal is still to find if the rule exist in the list.

pkg/cloudprovider/cloudapi/azure/azure_nsg_rules.go
```diff
@@ -51,63 +52,65 @@ func getDefaultDenyRuleName() string {
...
-func updateSecurityRuleNameAndPriority(existingRules []armnetwork.SecurityRule,
-	newRules []armnetwork.SecurityRule) []armnetwork.SecurityRule {
-	var rules []armnetwork.SecurityRule
-	defaultAndUserCreatedRules := make([]armnetwork.SecurityRule, 0)
-
+// updateSecurityRuleNameAndPriority updates rule name and priority for new security rules based on existing security rules and returns all.
+func updateSecurityRuleNameAndPriority(existingRules []*armnetwork.SecurityRule,
+	newRules []*armnetwork.SecurityRule) []*armnetwork.SecurityRule {
+	var rules []*armnetwork.SecurityRule
+	existingRulePriority := make(map[int32]struct{})
```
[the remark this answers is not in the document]
shenmo3: Here I put all variable declarations together, wouldn't a new line in between less ideal?

pkg/cloudprovider/cloudapi/azure/azure_nsg_rules.go
```diff
@@ -51,63 +52,65 @@ func getDefaultDenyRuleName() string {
...
-		rule.Name = &ruleName
-		rule.Properties.Priority = to.Int32Ptr(rulePriority)
-
 		rules = append(rules, rule)
-		rulePriority++
 	}
 
 	for _, rule := range newRules {
-		if rule.Properties == nil {
+		if rule == nil || rule.Properties == nil {
```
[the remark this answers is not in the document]
shenmo3: With the new change its not needed, removed.

pkg/cloudprovider/cloudapi/azure/azure_nsg_rules.go
```diff
@@ -51,63 +52,65 @@ func getDefaultDenyRuleName() string {
...
 			continue
 		}
 		// skip priority update for default 4096 deny rule. newRules consists of only Nephe rules.
 		if *rule.Properties.Priority == vnetToVnetDenyRulePriority {
-			defaultAndUserCreatedRules = append(defaultAndUserCreatedRules, rule)
+			rules = append(rules, rule)
 			continue
 		}
+
+		// find first unused priority.
```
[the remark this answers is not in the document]
shenmo3: It's just 5 lines and not being used anywhere else, I think moving it to a separate function is less readable. Would make sense if its used more than once.

pkg/cloudprovider/cloudapi/azure/azure_security.go
shenmo3: Yes. However this logic changes in the new revision

pkg/cloudprovider/cloudapi/azure/azure_nsg_rules.go
```diff
@@ -51,29 +52,23 @@ func getDefaultDenyRuleName() string {
...
-	newRules []armnetwork.SecurityRule) []armnetwork.SecurityRule {
-	var rules []armnetwork.SecurityRule
-	defaultAndUserCreatedRules := make([]armnetwork.SecurityRule, 0)
-
+// updateSecurityRuleNameAndPriority updates rule name and priority for new security rules based on existing security rules
+// and returns them combined.
+func updateSecurityRuleNameAndPriority(existingRules []*armnetwork.SecurityRule,
+	newRules []*armnetwork.SecurityRule) []*armnetwork.SecurityRule {
+	var rules []*armnetwork.SecurityRule
+	existingRulePriority := make(map[int32]struct{})
```
reachjainrahul, line 60: uint32?

pkg/cloudprovider/cloudapi/azure/azure_nsg_rules.go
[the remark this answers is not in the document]
reachjainrahul: better to move to new function.. its purpose is well defined.. looking at function call, while reading, we would know, that function will give us new unused priority, hence we dont have to look at the code/logic

pkg/cloudprovider/cloudapi/azure/azure_security.go
```diff
@@ -344,100 +336,172 @@ func (computeCfg *computeServiceConfig) buildEffectiveNSGSecurityRulesToApply(ap
...
 			_, _, isNepheControllerCreatedRule := securitygroup.IsNepheControllerCreatedSG(desc.AppliedToGroup)
-			if isNepheControllerCreatedRule && strings.Compare(ruleAddrGroupName, appliedToGroupNepheControllerName) == 0 {
-				continue
+			if isNepheControllerCreatedRule && strings.Compare(desc.AppliedToGroup, appliedToGroupNepheControllerName) == 0 {
+				// skip the rule if found in remove list.
+				removeRules := rmIngressSecurityRules
+				if *rule.Properties.Direction == armnetwork.SecurityRuleDirectionOutbound {
+					removeRules = rmEgressSecurityRules
+				}
+				idx, found := findSecurityRule(rule, removeRules)
```
reachjainrahul, line 381: here you are comparing full rules..In your internal CloudRuleIndexer, cannot you store priority or ID for AWS and just skip them for removal?

pkg/cloudprovider/cloudapi/azure/azure_security.go
```diff
@@ -344,100 +336,172 @@ func (computeCfg *computeServiceConfig) buildEffectiveNSGSecurityRulesToApply(ap
...
+	if nsgObj.Properties == nil {
+		return nil, fmt.Errorf("properties field empty in nsg object %s", *nsgObj.ID)
+	}
 
-	newIngressSecurityRules, err := convertIngressToPeerNsgSecurityRules(appliedToGroupID, ingressRules,
+	addIRule := make([]*securitygroup.CloudRule, 0)
+	rmIRule := make([]*securitygroup.CloudRule, 0)
+	addERule := make([]*securitygroup.CloudRule, 0)
+	rmERule := make([]*securitygroup.CloudRule, 0)
+	for _, rule := range addRules {
```
reachjainrahul, line 421: move to new function.. GetRules, pass ingress as boolean to figure out direction

pkg/cloudprovider/cloudapi/azure/azure_security.go
```diff
@@ -344,100 +336,172 @@ func (computeCfg *computeServiceConfig) buildEffectiveNSGSecurityRulesToApply(ap
...
 
-	agAsgMapByNepheControllerName, _, err := getNepheControllerCreatedAsgByNameForResourceGroup(computeCfg.asgAPIClient, rgName)
+	allIngressRules := updateSecurityRuleNameAndPriority(currentNsgIngressRules, addIngressSecurityRules)
+	allEgressRules := updateSecurityRuleNameAndPriority(currentNsgEgressRules, addEgressSecurityRules)
+
+	return append(allIngressRules, allEgressRules...), nil
+}
+
+// buildEffectivePeerNSGSecurityRulesToApply prepares the update rule cloud api payload from internal rules that require peering.
+func (computeCfg *computeServiceConfig) buildEffectivePeerNSGSecurityRulesToApply(appliedToGroupID *securitygroup.CloudResourceID,
```
reachjainrahul, line 405: We dont exercise/support peering code right, for non peer case

pkg/controllers/networkpolicy/controller_test.go
```diff
@@ -433,36 +432,60 @@ var _ = Describe("NetworkPolicy", func() {
...
+				currentIn = append(currentIn, *rule.Rule.(*securitygroup.IngressRule))
+			case *securitygroup.EgressRule:
+				currentEg = append(currentEg, *rule.Rule.(*securitygroup.EgressRule))
+			}
+		}
+		// number of current rules + number of adding rules - number of removing rules should equal to number of total target rules.
+		Expect(len(currentIn) + len(addIn) - len(rmIn)).To(Equal(len(ingressRule)))
+		Expect(len(currentEg) + len(addEg) - len(rmEg)).To(Equal(len(egressRule)))
+		// adding rules should not present in current rules and present it target rules.
+		for _, rule := range addIn {
```
reachjainrahul, line 472: dont use short name.. addIngress and currentIngress

pkg/cloudprovider/cloudapi/azure/azure_security.go
shenmo3: correct. We don't claim support for peering, its just that the peering code is here and I just update them too.

pkg/controllers/networkpolicy/controller_test.go
shenmo3: Done

pkg/cloudprovider/cloudapi/azure/azure_security.go
shenmo3: Done

pkg/cloudprovider/cloudapi/azure/azure_nsg_rules.go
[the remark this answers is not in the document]
shenmo3: Done

pkg/cloudprovider/cloudapi/azure/azure_security.go
shenmo3: Changed to normalize rule and use deepequal to simplify this logic. Indexer change is a larger change and should have a separate PR.

pkg/cloudprovider/cloudapi/azure/azure_security.go
```diff
@@ -297,21 +297,65 @@ func (computeCfg *computeServiceConfig) processAsgAttachDetachConcurrently(asgOb
...
+
+	addIRule, addERule := splitCloudRulesByDirection(addRules)
+	rmIRule, rmERule := splitCloudRulesByDirection(rmRules)
+
+	agAsgMapByNepheControllerName, atAsgMapByNepheControllerName, err := getNepheControllerCreatedAsgByNameForResourceGroup(
+		computeCfg.asgAPIClient, rgName)
+	if err != nil {
+		return []*armnetwork.SecurityRule{}, err
+	}
+	addIngressSecurityRules, err := convertIngressToNsgSecurityRules(appliedToGroupID, addIRule,
```
reachjainrahul, line 336: can you convert one function to get all the rules, and do error comparison just once.

pkg/controllers/networkpolicy/controller_test.go
```diff
@@ -433,36 +432,60 @@ var _ = Describe("NetworkPolicy", func() {
...
+			}
 		}
-		ipMap := make(map[string]*net.IPNet)
-		ipStrList := make([]string, 0)
-		for _, ip := range inputIP {
-			ipStrList = append(ipStrList, ip.String())
-			ipMap[ip.String()] = ip
+		rmIngress := make([]securitygroup.IngressRule, 0)
+		rmEgress := make([]securitygroup.EgressRule, 0)
+		for _, rule := range rm {
```
reachjainrahul, line 448: we have a new function for this right? splitCloudRulesByDirection

pkg/cloudprovider/cloudapi/azure/azure_security.go
shenmo3: I tried but I don't find a good way of doing it. Moving all these split and converts into a helper func works but the func signature is way too big and not readable. The convert function themselves are already complicated enough and I don't want to add more things into it. I shortened some names and reduced some lines here to help with cleaness.

pkg/controllers/networkpolicy/controller_test.go
shenmo3: Done

reachjainrahul (approved):
/LGTM

PR #226, opened 2023-05-16
change deprecated ginkgo flag --failFast to --fail-fast

shenmo3:
## Description
```
You're using deprecated Ginkgo functionality:
=============================================
  --failFast is deprecated, use --fail-fast instead
```

## Changes
1. Change --failFast to --fail-fast in Makefile


Anandkumar26 (approved):
/LGTM

PR #237, opened 2023-05-31
fix reconciliation ci error on controller restart

### LINKED ISSUE #236: Reconciliation CI test failure due to asg deletion
**Describe the bug**
CI failed on reconciliation test due to ANP change during controller restart. The address asg is deleted during syncFromCloud before updating rules, causing update rules unable to find asg when building remove rules, reporting an error.

**To Reproduce**
1. Import Azure VMs and apply ANP with address group.
1. Restart Nephe controller and modify the ANP's address group, before the controller fully restarted.

**Expected**
The old rule and asg is removed, new rule and asg applied.

**Actual behavior**
asg not found error reported.

**Versions:**
ToT

shenmo3:
## Description
This PR fixes #236. On rule addition, we want to check if target address asg exists and report an error if not. On deletion we don't care about this since the related rules must have been cleaned up on asg deletion. A check for deletion or not is added before reporting error.

## Changes
1. Add a parameter and check on building and converting Azure rules, so that asg not found will not report an error on rule deletion.

### DIFF
--- a/pkg/cloudprovider/cloudapi/azure/azure_nsg_rules.go
+++ b/pkg/cloudprovider/cloudapi/azure/azure_nsg_rules.go
@@ -121,7 +121,7 @@ func updateSecurityRuleNameAndPriority(existingRules []*armnetwork.SecurityRule,
 }
 
 // convertIngressToNsgSecurityRules converts ingress rules from securitygroup.CloudRule to azure rules.
-func convertIngressToNsgSecurityRules(appliedToGroupID *securitygroup.CloudResourceID, rules []*securitygroup.CloudRule,
+func convertIngressToNsgSecurityRules(appliedToGroupID *securitygroup.CloudResourceID, rules []*securitygroup.CloudRule, isRemove bool,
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/cloudapi/azure/azure_nsg_rules.go
```diff
@@ -162,7 +162,10 @@ func convertIngressToNsgSecurityRules(appliedToGroupID *securitygroup.CloudResou
 
 		srcApplicationSecurityGroups, err := convertToAzureApplicationSecurityGroups(rule.FromSecurityGroups, agAsgMapByNepheControllerName)
 		if err != nil {
-			return []*armnetwork.SecurityRule{}, err
+			if !isRemove {
```
reachjainrahul, line 165: lets revisit this. dont like the idea of passing flag in convert function. Allowing this checkin for now

reachjainrahul (approved):
/LGTM

PR #238, opened 2023-06-01
Upgrade antrea version to v1.12

reachjainrahul:


pkg/apiserver/registry/inventory/virtualmachine/rest.go
```diff
@@ -64,6 +64,9 @@ func (r *REST) New() runtime.Object {
 	return &runtimev1alpha1.VirtualMachine{}
 }
 
+func (r *REST) Destroy() {
```
shenmo3, line 67: Is this a new requirement for upgrading?

shenmo3 (approved):
/LGTM with a minor question.

pkg/apiserver/registry/inventory/virtualmachine/rest.go
reachjainrahul: Yes.. with antrea 1.12, runtime version is upgraded too.

shenmo3 (approved):

PR #244, opened 2023-06-05
stop converting azure any rule to ipv4 cidr

shenmo3:
## Description
Currently, Azure's "any" rule includes both IPv4 and IPv6. While converting the "any" rule into an IPv4 CIDR makes sense when only IPv4 is supported, it becomes incorrect if IPv6 is also taken into picture.

## Change
1. Eliminated the special handling that converts Azure's "any" rule into an IPv4 0.0.0.0/0 CIDR. From now on, the "any" rule in Azure will be treated as a user-defined custom rule and will not be converted by Nephe. Nephe will only handle specific IPv4 and IPv6 CIDRs.


reachjainrahul (approved):
/LGTM

PR #246, opened 2023-06-07
Adding IPv6 rule support in Nephe

reachjainrahul:
## Description
Handle IPv6 addresses configured on Antrea NetworkPolicy

## Changes 
With this checkin, Nephe controller will accept IPv6 IP blocks in Antrea NetworkPolicy and will apply them in cloud. Also, VirtualMachine Object will now include both IPv4 and IPv6 addresses in the network interface field.

### DIFF
--- a/pkg/cloudprovider/cloudapi/aws/aws_converters.go
+++ b/pkg/cloudprovider/cloudapi/aws/aws_converters.go
@@ -45,28 +45,44 @@ func convertToIPPermissionPort(port *int, protocol *int) (*int64, *int64) {
 	return portVal, portVal
 }
 
-func convertToEc2IpRanges(ips []*net.IPNet, ruleHasGroups bool, description *string) []*ec2.IpRange {
-	var ipRanges []*ec2.IpRange
+// convertToEc2IpRanges convert the IPs to EC2 IPv4 and IPv6 ranges.
+func convertToEc2IpRanges(ips []*net.IPNet, ruleHasGroups bool, description *string) ([]*ec2.IpRange, []*ec2.Ipv6Range) {
+	var ipv4Ranges []*ec2.IpRange
+	var ipv6Ranges []*ec2.Ipv6Range
 	if len(ips) == 0 && !ruleHasGroups {
 		ipRange := &ec2.IpRange{
 			CidrIp:      aws.String("0.0.0.0/0"),
 			Description: description,
 		}
-		ipRanges = append(ipRanges, ipRange)
-		return ipRanges
+		ipv4Ranges = append(ipv4Ranges, ipRange)
+		ipv6Range := &ec2.Ipv6Range{
+			CidrIpv6:    aws.String("::/0"),
+			Description: description,
+		}
+		ipv6Ranges = append(ipv6Ranges, ipv6Range)
+		return ipv4Ranges, ipv6Ranges
 	}
 
 	for _, ip := range ips {
-		ipRange := &ec2.IpRange{
-			CidrIp:      aws.String(ip.String()),
-			Description: description,
+		if ip.IP.To4() != nil {
+			ipRange := &ec2.IpRange{
+				CidrIp:      aws.String(ip.String()),
+				Description: description,
+			}
+			ipv4Ranges = append(ipv4Ranges, ipRange)
+		} else if ip.IP.To16() != nil {
+			ipRange := &ec2.Ipv6Range{
+				CidrIpv6:    aws.String(ip.String()),
+				Description: description,
+			}
+			ipv6Ranges = append(ipv6Ranges, ipRange)
 		}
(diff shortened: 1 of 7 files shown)


reachjainrahul:
/nephe-test-e2e-azure, /nephe-test-e2e-aks /nephe-test-e2e-eks

pkg/cloudprovider/cloudapi/aws/aws_crd_helpers.go
```diff
@@ -46,30 +46,41 @@ func ec2InstanceToInternalVirtualMachineObject(instance *ec2.Instance, vpcs map[
...
 		}
+		// Parsing Amazon allocated IPv6 addresses for now. They are treated as Public IPs.
+		// TODO: Re-check the parsing for private IPv6 IP block.
+		if len(nwInf.Ipv6Addresses) > 0 {
+			for _, ipv6Address := range nwInf.Ipv6Addresses {
+				ipAddress := runtimev1alpha1.IPAddress{
+					AddressType: runtimev1alpha1.AddressTypeExternalIP,
+					Address:     *ipv6Address.Ipv6Address,
+				}
+				ipAddressObjs = append(ipAddressObjs, ipAddress)
```
Anandkumar26, line 77: For the agented, are the Externalnodes getting created correctly?

pkg/cloudprovider/cloudapi/aws/aws_converters.go
```diff
@@ -45,28 +45,39 @@ func convertToIPPermissionPort(port *int, protocol *int) (*int64, *int64) {
...
+	var ipv6Ranges []*ec2.Ipv6Range
 	if len(ips) == 0 && !ruleHasGroups {
 		ipRange := &ec2.IpRange{
 			CidrIp:      aws.String("0.0.0.0/0"),
 			Description: description,
 		}
-		ipRanges = append(ipRanges, ipRange)
-		return ipRanges
+		ipv4Ranges = append(ipv4Ranges, ipRange)
+		return ipv4Ranges, ipv6Ranges
```
shenmo3, line 58: This means allow all right? So shouldn't an IPv6 `::` range also be added?

pkg/cloudprovider/cloudapi/aws/aws_crd_helpers.go
```diff
@@ -46,30 +46,41 @@ func ec2InstanceToInternalVirtualMachineObject(instance *ec2.Instance, vpcs map[
...
 					}
-					ipAddressCRDs = append(ipAddressCRDs, ipAddressCRD)
+					ipAddressObjs = append(ipAddressObjs, ipAddressCRD)
 				}
 			}
 		}
+		// Parsing Amazon allocated IPv6 addresses for now. They are treated as Public IPs.
+		// TODO: Re-check the parsing for private IPv6 IP block.
+		if len(nwInf.Ipv6Addresses) > 0 {
+			for _, ipv6Address := range nwInf.Ipv6Addresses {
```
shenmo3, line 72: is the if block needed? a for block is sufficient isn't it?

pkg/cloudprovider/cloudapi/aws/aws_crd_helpers.go
reachjainrahul: Yes..

pkg/inventory/inventory.go
```diff
@@ -248,7 +250,35 @@ func (inventory *Inventory) GetVmByKey(key string) (*runtimev1alpha1.VirtualMach
...
+
+// compareVirtualMachineObjects compare if two virtual machine objects are the same. Return true if same.
+func (i Inventory) compareVirtualMachineObjects(cached, discovered runtimev1alpha1.VirtualMachineStatus) bool {
+	// 1. Check if objects are same.
+	if reflect.DeepEqual(cached, discovered) {
+		return true
+	}
+	// 2. Check if NetworkInterface field differ.
+	if reflect.DeepEqual(cached.NetworkInterfaces, discovered.NetworkInterfaces) {
+		return false
```
shenmo3, line 266: shouldn't this return true if nics are same?

reachjainrahul: Azure returns nic address not in particular order. So we sort them and then compare

pkg/cloudprovider/cloudapi/aws/aws_converters.go
reachjainrahul: you are right.. let me add that

pkg/cloudprovider/cloudapi/aws/aws_crd_helpers.go
reachjainrahul: its generally a good practice, but not must..

shenmo3 (approved):

PR #247, opened 2023-06-07
fix reconciliation issue when ag is modified during controller restart

shenmo3:
## Description
Currently, if the address group of network policies is modified during controller restart, the reconciliation process will encounter issues. The syncFromCloud will delete the previous address group before deleting the previous rules that still reference it. This lead to dependency violation in AWS and ASG not found error during rule updates in Azure. This PR addresses the problem by first updating address group references for appliedTo groups before deleting address groups.

## Changes
1. The appliedTo group sync now computes the address group references before performing rule updates on indexer changes, instead of only updating references after the async rule update completes.
2. SyncFromCloud will delete unknown address groups after appliedTo group sync finishes.
3. The isRemove flag, previously added for Azure to fix the ASG not found issue, has been removed.

### DIFF
--- a/pkg/cloudprovider/cloudapi/azure/azure_security.go
+++ b/pkg/cloudprovider/cloudapi/azure/azure_security.go
@@ -316,19 +316,19 @@ func (computeCfg *computeServiceConfig) buildEffectiveNSGSecurityRulesToApply(ap
 	if err != nil {
 		return []*armnetwork.SecurityRule{}, err
 	}
-	addIngressRules, err := convertIngressToNsgSecurityRules(appliedToGroupID, addIRule, false, agAsgMapByNepheName, atAsgMapByNepheName)
+	addIngressRules, err := convertIngressToNsgSecurityRules(appliedToGroupID, addIRule, agAsgMapByNepheName, atAsgMapByNepheName)
 	if err != nil {
 		return []*armnetwork.SecurityRule{}, err
 	}
-	addEgressRules, err := convertEgressToNsgSecurityRules(appliedToGroupID, addERule, false, agAsgMapByNepheName, atAsgMapByNepheName)
+	addEgressRules, err := convertEgressToNsgSecurityRules(appliedToGroupID, addERule, agAsgMapByNepheName, atAsgMapByNepheName)
 	if err != nil {
 		return []*armnetwork.SecurityRule{}, err
 	}
-	rmIngressRules, err := convertIngressToNsgSecurityRules(appliedToGroupID, rmIRule, true, agAsgMapByNepheName, atAsgMapByNepheName)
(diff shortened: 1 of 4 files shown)


pkg/cloudprovider/cloudapi/azure/azure_security.go
```diff
@@ -1020,9 +1020,12 @@ func (c *azureCloud) DeleteSecurityGroup(securityGroupIdentifier *securitygroup.
 	if err != nil {
 		return err
 	}
-	err = computeService.removeReferencesToSecurityGroup(&securityGroupIdentifier.CloudResourceID, rgName, location, membershipOnly)
-	if err != nil {
-		return err
+	// remove attached rules for appliedTo sg but not address sg. this behavior is consistent with AWS.
```
reachjainrahul, line 1023: This

pkg/controllers/networkpolicy/sync.go
```diff
@@ -197,23 +198,29 @@ func (r *NetworkPolicyReconciler) syncWithCloud() {
...
+		if content.MembershipOnly {
+			// mark unknown address groups and pending delete groups for deletion after appliedTo groups updates address group references.
+			obj, ok, _ := r.addrSGIndexer.GetByKey(content.Resource.CloudResourceID.String())
+			if !ok {
+				state := securityGroupStateCreated
+				sg := newAddrSecurityGroup(&content.Resource, []*securitygroup.CloudResource{}, &state).(*addrSecurityGroup)
+				removeAddrSgs = append(removeAddrSgs, sg)
+				continue
+			}
+			if obj.(*addrSecurityGroup).deletePending {
```
reachjainrahul, line 213: Put this in else if block

pkg/controllers/networkpolicy/sync.go
```diff
@@ -162,6 +162,7 @@ func (a *appliedToSecurityGroup) sync(syncContent *securitygroup.Synchronization
 	}
 
 	if indexerUpdate {
+		_ = a.updateAddrGroupReference(r)
```
reachjainrahul, line 165: How is the old address group reference added to AT?

shenmo3: The old rule referencing previous address group is synced into the indexer for the current AT, and have an indexer update, so this func call will update the reference in AT.

pkg/cloudprovider/cloudapi/azure/azure_security.go
shenmo3: Done

pkg/controllers/networkpolicy/sync.go
shenmo3: Done

pkg/controllers/networkpolicy/sync.go
shenmo3: Done

PR #248, opened 2023-06-08
Set poll interval is not specified in cpa cr

reachjainrahul:
## Description
Fix nephe controller crash if poll interval is not specified and webhook is disabled.

## Changes
Add a check to see if the field is not specified, explicitly set to default 60 sec.


Anandkumar26 (approved):
/LGTM

PR #250, opened 2023-06-14
Fix watch handler for vpc and vm objects

reachjainrahul:
## Description
For both vpc and vm objects watch, the selectors were not updated with
- Namespace: if not specified in field selector. Filter like -n <ns> was not working with both label and field selector.
- Key: namespace in the key was not updated leading to failure on key based watch. Primary key to cache is namespaced/name.
- Fields: like status.region, status.cloudId was not set.

## Changes
Fix selector to include namespace/name as key and update namespace in metadata.namespace field of selector if specific namespace is specified. Further, new field selector are updated in inventory store of vm and vpc.
Also updated the list call of vm and vpc to first include all objects if -A (all namespaces) is specified and then filter objects with labels and filter.

### DIFF
--- a/pkg/apiserver/registry/inventory/virtualmachine/rest.go
+++ b/pkg/apiserver/registry/inventory/virtualmachine/rest.go
@@ -26,13 +26,15 @@ import (
 	metatable "k8s.io/apimachinery/pkg/api/meta/table"
 	"k8s.io/apimachinery/pkg/apis/meta/internalversion"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/apiserver/pkg/registry/rest"
 
 	runtimev1alpha1 "antrea.io/nephe/apis/runtime/v1alpha1"
-	registryinventory "antrea.io/nephe/pkg/apiserver/registry/inventory"
+	"antrea.io/nephe/pkg/apiserver/registry/inventory/selector"
+	"antrea.io/nephe/pkg/apiserver/registry/inventory/util"
 	"antrea.io/nephe/pkg/inventory"
 	"antrea.io/nephe/pkg/inventory/indexer"
 	"antrea.io/nephe/pkg/inventory/store"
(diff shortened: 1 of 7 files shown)


pkg/inventory/store/vm_inventory.go
```diff
@@ -49,20 +49,28 @@ func keyAndSpanSelectFuncVm(selectors *antreastorage.Selectors, key string, obj
 	if selectors.Key != "" && key != selectors.Key {
 		return false
 	}
+	if selectors.Label.Empty() && selectors.Field.Empty() {
+		// Match everything.
```
Anandkumar26, line 53: This scenario corresponds `kubectl get vm -A -w`?

pkg/apiserver/registry/inventory/virtualmachine/rest.go
```diff
@@ -164,9 +170,40 @@ func (r *REST) ConvertToTable(_ context.Context, obj runtime.Object, _ runtime.O
...
+	contextNamespace, _ := request.NamespaceFrom(ctx)
+	if key != "" {
+		if contextNamespace == "" {
+			return nil, errors.NewBadRequest("cannot watch for a resources in all namespaces")
+		} else {
+			key = contextNamespace + "/" + key
+		}
+	} else if contextNamespace != "" {
+		// Namespace specified and field selector doesn't have namespace.
+		if _, ok := field.RequiresExactMatch("metadata.namespace"); !ok {
```
Anandkumar26, line 182: Why do we need explicit handling for metadata.namespace?

The keyAndSpanSelectFuncVm is anyways going to match the object against with fieldselectors and labelselectors.

Anandkumar26 (commented):
Overall it looks good to me.

pkg/inventory/store/vm_inventory.go
reachjainrahul: Yes

pkg/apiserver/registry/inventory/virtualmachine/rest.go
reachjainrahul: If -n is specified, I set field selector only when field selector doesn't have metadata.namespace set. If I don't have the check then -n ns1 --field-selector metadata.namespace=ns2, ns2 will be overwritten. In this case I am giving ns2 priority over ns1. Although it's a invalid case.

Anandkumar26 (approved):
/LGTM

PR #252, opened 2023-06-15
fix race condition when rule level appliedTo groups are configured

shenmo3:
## Description
Currently, when address group becomes ready after network policy and appliedTo group are ready, then the `notifyAddressGroupChange` func will only trigger update rules for policy-level appliedTo groups, while ignoring rule-level appliedTo groups. This PR resolves this problem by adding triggers for all policy-level and rule-level appliedTo groups.

## Changes
1. Modified the `notifyAddressGroupChange` function to utilize the `getAppliedToGroups` method, which retrieves both policy-level and rule-level appliedTo groups for triggering rule updates.
2. Optimized the `getAppliedToGroups` method to return only unique appliedTo groups,


reachjainrahul (approved):
/LGTM

PR #259, opened 2023-06-22
Error handling for AppliedToGroup SG delete.

Anandkumar26:
Consider a scenario where ANP contains two rule-level AppliedToGroups. Delete one of the AppliedToGroup. If the ATGroup could not be deleted on the cloud, then the internal state of NP controller still contains the rules, while the rules are deleted on the cloud.

This PR tries to address the issue, by clearing the internal state when an ATGroup is deleted, so that if the ATGroup is re-added, then NP controller will be able to recover and realize the rules on the cloud.

### DIFF
--- a/pkg/controllers/networkpolicy/networkpolicy.go
+++ b/pkg/controllers/networkpolicy/networkpolicy.go
@@ -421,7 +421,7 @@ func (s *securityGroupImpl) deleteImpl(c cloudSecurityGroup, membershipOnly bool
 			var err error
 			nps, err = r.networkPolicyIndexer.ByIndex(networkPolicyIndexerByAppliedToGrp, s.id.Name)
 			if err != nil {
-				return fmt.Errorf("get networkpolicy indexer with index=%v, name=%v: %v", networkPolicyIndexerByAppliedToGrp, s.id.Name, err)
+				return fmt.Errorf("failed to get from networkPolicy indexer %v with key %v: %w", networkPolicyIndexerByAppliedToGrp, s.id.Name, err)
 			}
 		}
 		if err := indexer.Delete(c); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/controllers/networkpolicy/networkpolicy.go
```diff
@@ -426,9 +427,6 @@ func (s *securityGroupImpl) deleteImpl(c cloudSecurityGroup, membershipOnly bool
 	ch := securitygroup.CloudSecurityGroup.DeleteSecurityGroup(&s.id, membershipOnly)
 	go func() {
 		err := <-ch
-		if err == nil && !membershipOnly {
-			s.deleteSgRulesFromIndexer(r)
```
reachjainrahul, line 430: why you dont want to remove the rules from indexer for success case?

pkg/controllers/networkpolicy/networkpolicy.go
```diff
@@ -1397,6 +1393,7 @@ func (n *networkPolicy) update(anp *antreanetworking.NetworkPolicy, recompute bo
 	modifiedAddr := make([]string, 0)
 	if recompute {
 		addedAddr, _ = diffAddressGrp(nil, anp.Rules)
+		r.Log.V(1).Info("Compute rules", "networkPolicy", n.Name)
```
reachjainrahul, line 1396: too many recompute logs.. 1396, 1418, 1440. Just have logs in computeRules func only..

pkg/controllers/networkpolicy/networkpolicy.go
Anandkumar26: Discussed offline and updated PR.

pkg/controllers/networkpolicy/networkpolicy.go
Anandkumar26: These logs makes it easier to understand np controller workflow.

reachjainrahul (approved):
/LGTM

PR #260, opened 2023-06-22
Move Plugin level lock to Account level

reachjainrahul:
## Description
We had mutex lock at the plugin level, due to which any cloud operation for account1 was impacting account2. 

## Changes
Move plugin level lock to account level so that multiple cloud operation can happen in parallel. There is further room for optimization to identify what cloud calls can happen in parallel within an account rather than locking at account level.

### DIFF
--- a/pkg/cloudprovider/plugins/azure/azure_security_impl.go
+++ b/pkg/cloudprovider/plugins/azure/azure_security_impl.go
@@ -19,22 +19,19 @@ import (
 	"fmt"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork"
 	"github.com/Azure/go-autorest/autorest/to"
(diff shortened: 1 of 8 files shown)


pkg/cloudprovider/plugins/azure/azure_security_impl.go
```diff
@@ -17,31 +17,29 @@ package azure
...
 	"context"
 	"fmt"
-	"strings"
-	"sync"
-	"time"
-
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork"
 	"github.com/Azure/go-autorest/autorest/to"
 	"k8s.io/apimachinery/pkg/types"
+	"strings"
```
archanapholla, line 23: could you sort imports here?

reachjainrahul: Done

archanapholla (approved):
/LGTM

PR #262, opened 2023-06-26
Add support to clean Np indexers based on account

Anandkumar26:
- When CPA is deleted, it will notify a local event to np controller.
- As part of processing local event, np controller will clear entries corresponding to the account from appliedToSGIndexer, addrSGIndexer, cloudRuleIndexer and networkPolicyIndexer.
- Remove AT/AG groups from retryQueue for the account.
- Remove groups from pendingDeleteGroups based on account.

### DIFF
--- a/pkg/controllers/networkpolicy/controller.go
+++ b/pkg/controllers/networkpolicy/controller.go
@@ -35,6 +35,7 @@ import (
 	antreav1alpha1 "antrea.io/antrea/pkg/apis/crd/v1alpha1"
 	antreav1alpha2 "antrea.io/antrea/pkg/apis/crd/v1alpha2"
 	antreanetworkingclient "antrea.io/antrea/pkg/client/clientset/versioned/typed/controlplane/v1beta2"
+	crdv1alpha1 "antrea.io/nephe/apis/crd/v1alpha1"
 	"antrea.io/nephe/pkg/cloudprovider/cloudresource"
 	"antrea.io/nephe/pkg/cloudprovider/securitygroup"
 	"antrea.io/nephe/pkg/config"
@@ -45,6 +46,7 @@ import (
 const (
 	NetworkPolicyStatusIndexerByNamespace       = "namespace"
 	addrAppliedToIndexerByGroupID               = "GroupID"
+	addrAppliedToIndexerByAccountId             = "AccountId"
 	appliedToIndexerByAddrGroupRef              = "AddressGrp"
 	networkPolicyIndexerByAddrGrp               = "AddressGrp"
 	networkPolicyIndexerByAppliedToGrp          = "AppliedToGrp"
@@ -56,7 +58,7 @@ const (
 	cloudResponseChBuffer = 50
 
 	// NetworkPolicy controller is ready to sync after it receives bookmarks from
-	// networkpolicy, addrssGroup and appliedToGroup.
+	// networkpolicy, addressGroup and appliedToGroup.
 	npSyncReadyBookMarkCnt = 3
 )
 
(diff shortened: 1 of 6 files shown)


pkg/controllers/networkpolicy/controller.go
```diff
@@ -466,7 +468,9 @@ func (r *NetworkPolicyReconciler) processLocalEvent(event watch.Event) error {
 		return r.processAppliedToGroup(event)
 	case *antreanetworking.AppliedToGroupPatch:
 		return r.processAppliedToGroup(event)
-
+	case *crdv1alpha1.CloudProviderAccount:
+		cpa := event.Object.(*crdv1alpha1.CloudProviderAccount)
+		return r.RemoveIndexerObjectsByAccount(types.NamespacedName{Name: cpa.Name, Namespace: cpa.Namespace}.String())
```
reachjainrahul, line 473: nit remove

pkg/controllers/networkpolicy/controller.go
```diff
@@ -764,3 +782,83 @@ func (r *NetworkPolicyReconciler) resetWatchers() error {
...
+	atSgs, err := r.appliedToSGIndexer.ByIndex(addrAppliedToIndexerByAccountId, namespacedName)
+	if err != nil {
+		r.Log.Error(err, "failed to get appliedToGroups from indexer", "account", namespacedName)
+	}
+	if len(atSgs) == 0 {
+		r.Log.V(1).Info("No appliedToGroups found for account", "account", namespacedName)
+	}
+	for i := range atSgs {
+		if atSgs[i] == nil {
+			continue
```
reachjainrahul, line 799: Why would this be nil?

Anandkumar26: While testing out namespace deletion, I observed a crash saying `interface is nil`.
I observed it while traversing agSgs.

Anandkumar26: When we retrieve objects using by Index, the length of the objects returned can be non-zero, yet while accessing the obj(interface), the obj(interface) can be nil. 

I tried this by deleting an object in the indexer and got all the objects using Byindex immediately. 
one of the object returned was nil.

So while retrieving objects by Index, we should check if the interface can be converted to a required object and then proceed.

pkg/controllers/networkpolicy/controller.go
```diff
@@ -764,3 +782,95 @@ func (r *NetworkPolicyReconciler) resetWatchers() error {
...
+		for _, obj := range agSgs {
+			agSg, ok := obj.(*addrSecurityGroup)
+			if !ok {
+				continue
+			}
+
+			// Remove AG group from retryQueue and pendingDeleteGroups.
+			uName := getGroupUniqueName(agSg.id.CloudResourceID.String(), true)
+			uGroupName := getGroupUniqueName(agSg.id.Name, true)
+			r.retryQueue.Remove(uName)
```
reachjainrahul, line 866: This wont work. In case of delete, we remove the item from indexer first, and then put in retry queue. So, you wont find the item for lookup in retry queue. You might have to iterate over items in the queue and delete all matching account-id.

pkg/controllers/networkpolicy/controller.go
```diff
@@ -764,3 +782,111 @@ func (r *NetworkPolicyReconciler) resetWatchers() error {
...
+
+			r.Log.V(1).Info("Deleting addressGroup from indexer", "addressGroup", agSg.id.Name)
+			if err := r.addrSGIndexer.Delete(agSg); err != nil {
+				r.Log.Error(err, "failed to delete addressGroup from indexer", "addressGroup", agSg.id.Name)
+			}
+		}
+	}
+
+	// Remove AT and AG group from retryQueue.
+	for _, item := range r.retryQueue.items {
```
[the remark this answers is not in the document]
Anandkumar26: In PendingGroup, we only store the GroupMember, so we wont be able to delete an entry based on account.

https://github.com/antrea-io/nephe/pull/262/files#diff-e370a727c5b2564b58693dfd2ff881f7df935871c0c0c806047c6613c4549c97R246

pkg/controllers/networkpolicy/controller.go
```diff
@@ -764,3 +782,119 @@ func (r *NetworkPolicyReconciler) resetWatchers() error {
...
+					r.retryQueue.Remove(uName)
+					break
+				}
+			}
+		}
+	}
+
+	for _, item := range r.pendingDeleteGroups.items {
+		group, ok := item.PendingItem.(*pendingGroup)
+		if ok && group.account == namespacedName {
```
reachjainrahul, line 893: nit: Rename to accountID for next PR, to be consistent with Retry Queue

reachjainrahul (approved):
/LGTM

PR #263, opened 2023-06-26
Track appliedTo group to network policies

reachjainrahul:
## Description
Tracker is used to update NP Status on Virtual Machine Policy Object. In an event ANP is deleted and cloud resources cannot be cleaned up, the ANP information is lost as its deleted from indexer leading to misleading info in vmp cache.

## Changes
With this PR, we store the mapping of appliedTo to Network Policy and use it for cases when ANP is not available in the cache.


Anandkumar26 (approved):
/LGTM

PR #265, opened 2023-06-27
Change cert and config location in manifest

reachjainrahul:
## Description
Both Webhook and API Server are using same TLS cert generated by cert manager. Avoid generating two volumes and different mount path for them

## Chanages
Move config to /etc/nephe dir and also update API Server and Webhook to use cert from same location.


reachjainrahul:
nephe-test-e2e-agentless-upgrade

Anandkumar26 (approved):
/LGTM

PR #269, opened 2023-07-05
Restart poller after all CES are reconciled after nephe restart.

archanapholla:


pkg/accountmanager/manager.go
```diff
@@ -61,6 +61,8 @@ type AccountManager struct {
 	Inventory        inventory.Interface
 	accPollers       map[types.NamespacedName]*accountPoller
 	accountConfigMap map[types.NamespacedName]*accountConfig
+	// accountToNumSelectors stores number of configured CloudEntitySelector CRs for a given account, required for handling nephe restart case.
```
Anandkumar26, line 64: nit: split comment into two lines.

pkg/accountmanager/manager.go
```diff
@@ -159,31 +164,56 @@ func (a *AccountManager) RemoveAccount(namespacedName *types.NamespacedName) err
...
+		}
+	}()
+
 	cloudProviderType, ok := a.getAccountProviderType(accNamespacedName)
 	if !ok {
-		return true, fmt.Errorf(fmt.Sprintf("failed to add or update selector %v, account %v: "+
-			"provider type not found", selectorNamespacedName, accNamespacedName))
+		retError = fmt.Errorf(fmt.Sprintf("failed to add or update selector %v, account %v: "+
+			"provider type not found", *selectorNamespacedName, *accNamespacedName))
+		return false, retError
```
Anandkumar26, line 192: What is the use of creating a new error variable here vs directly returning the error?

pkg/apiserver/webhook/cloudentityselector_webhook.go
```diff
@@ -83,10 +83,10 @@ func (v *CESMutator) Handle(_ context.Context, req admission.Request) admission.
 	referencedAccount := &v1alpha1.CloudProviderAccount{}
 	err = v.Client.Get(context.TODO(), *accountNameSpacedName, referencedAccount)
```
Anandkumar26, line 84: nit: Change to if err = (); err != nil {}

pkg/cloudprovider/plugins/aws/aws_ec2.go
```diff
@@ -127,7 +127,7 @@ func (ec2Cfg *ec2ServiceConfig) getManagedVpcIDs() map[string]struct{} {
 		awsPluginLogger().V(4).Info("Cache snapshot nil", "type", providerType, "account", ec2Cfg.accountNamespacedName)
 		return vpcIDsCopy
 	}
-	vpcIDsSet := snapshot.(*ec2ResourcesCacheSnapshot).managedVpcIDs
+	vpcIDsSet := snapshot.(*ec2ResourcesCacheSnapshot).managedVpcIds
```
Anandkumar26, line 130: nit: vpcIdsSet, vpcIdsCopy. 
If you are renaming the variables, please rename all the above.

pkg/controllers/cloudentityselector/controller.go
```diff
@@ -69,11 +69,10 @@ func (r *CloudEntitySelectorReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, err
```
Anandkumar26, line 69: nit:  return ctrl.Result{}, r.processDelete(&req.NamespacedName) and remove L68

pkg/accountmanager/manager.go
archanapholla: we want the defer function to get executed too which can alter the return value.

pkg/accountmanager/manager.go
```diff
@@ -159,31 +164,56 @@ func (a *AccountManager) RemoveAccount(namespacedName *types.NamespacedName) err
...
+			pendingCes--
+			a.accountToNumSelectors[*accNamespacedName] = pendingCes
+		}
+		if !found || pendingCes == 0 {
+			delete(a.accountToNumSelectors, *accNamespacedName)
+			accPoller, exists := a.getAccountPoller(accNamespacedName)
+			if exists {
+				accPoller.restartPoller(accNamespacedName)
+				// wait for polling to complete after restart.
+				retError = accPoller.waitForPollDone(accNamespacedName)
```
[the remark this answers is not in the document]
archanapholla: Error from "accPoller.waitForPollDone(accNamespacedName)" is not lost as we can have this kind of wait function in defer function. 
However, I saw an issue setting return value of AddResourceFiltersToAccount() from defer function.
I had to return the address.
Also, I tested the negative scenario, we should be good.

pkg/accountmanager/manager.go
archanapholla: Done

pkg/apiserver/webhook/cloudentityselector_webhook.go
archanapholla: Done

pkg/cloudprovider/plugins/aws/aws_ec2.go
archanapholla: Fixed all in azure_compute.go and aws_ec2.go .

pkg/accountmanager/manager.go
```diff
@@ -158,32 +164,57 @@ func (a *AccountManager) RemoveAccount(namespacedName *types.NamespacedName) err
 // AddResourceFiltersToAccount add/update cloud plugin for a given account to include the new selector and
 // restart poller once done.
 func (a *AccountManager) AddResourceFiltersToAccount(accNamespacedName *types.NamespacedName,
-	selectorNamespacedName *types.NamespacedName, selector *crdv1alpha1.CloudEntitySelector, replay bool) (bool, error) {
+	selectorNamespacedName *types.NamespacedName, selector *crdv1alpha1.CloudEntitySelector, replay bool) (bool, *error) {
```
reachjainrahul, line 167: Why is error used as pointer.. Consider changing it

pkg/accountmanager/manager.go
```diff
@@ -102,11 +106,13 @@ func (a *AccountManager) AddAccount(namespacedName *types.NamespacedName, accoun
 	// Create an account poller for polling cloud inventory.
 	accPoller, exists := a.addAccountPoller(cloudInterface, namespacedName, account)
 	if !exists {
-		if !crd.DoesCesCrExistsForAccount(a.Client, namespacedName) {
+		numCes := crd.GetNumberOfCesCrForAccount(a.Client, namespacedName)
```
reachjainrahul, line 109: GetCesCrForAccount Return CES and Count.. In the caller avoid using CES and just use count whereever possible

pkg/accountmanager/manager.go
archanapholla: Without it being a pointer, we cannot set return value in defer function. If it is a regular variable then the retError set in defer function doesn't get returned to the caller. As a side-effect status setting in CES wouldn't work.

pkg/controllers/cloudentityselector/controller.go
```diff
@@ -138,13 +136,13 @@ func (r *CloudEntitySelectorReconciler) processCreateOrUpdate(selector *crdv1alp
 		Name:      selector.Spec.AccountName,
 	}
 	r.selectorToAccountMap[*selectorNamespacedName] = *accountNamespacedName
-	retry, err := r.AccManager.AddResourceFiltersToAccount(accountNamespacedName, selectorNamespacedName,
+	// No errors can be retried for a CES.
+	_, err := r.AccManager.AddResourceFiltersToAccount(accountNamespacedName, selectorNamespacedName,
```
reachjainrahul, line 140: @Anandkumar26 Can you confirm this?

Anandkumar26: Yes. The error while processing CES can only occur if an account is either not initialised successfully or account is in a bad state.  I don't think there is any benefit in retrying the errors forever. 

So we capture the error and store in a CES status field. (which can be replayed when account goes from an uninitialised to initialised state OR it would need a user intervention)

pkg/controllers/cloudentityselector/controller.go
```diff
@@ -144,6 +143,12 @@ func (r *CloudEntitySelectorReconciler) processCreateOrUpdate(selector *crdv1alp
 		return err
 	}
 	r.updateStatus(selectorNamespacedName, err)
+	// Make sure reconciler retriable error cannot occur during restart, it can affect the logic in UpdatePendingCes function.
+	if count := r.AccManager.UpdatePendingCesCount(accountNamespacedName); count == 0 {
+		if err := r.AccManager.WaitForPollDone(accountNamespacedName); err != nil {
+			r.Log.Error(err, "Inventory poll failed", "account", accountNamespacedName, "error", err)
```
reachjainrahul, line 149: nit: inventory

pkg/controllers/cloudentityselector/controller.go
```diff
@@ -144,6 +143,12 @@ func (r *CloudEntitySelectorReconciler) processCreateOrUpdate(selector *crdv1alp
 		return err
 	}
 	r.updateStatus(selectorNamespacedName, err)
+	// Make sure reconciler retriable error cannot occur during restart, it can affect the logic in UpdatePendingCes function.
```
reachjainrahul, line 146: add detail comment on what it means..

archanapholla: added a TODO as discussed

pkg/controllers/cloudentityselector/controller.go
archanapholla: done

reachjainrahul (approved):
/LGTM

PR #272, opened 2023-07-06
Allow user tags on virtual machine

reachjainrahul:
## Description
In addition to discovered tags from cloud, we now allow user tags on virtual machine objects. These user tags will also be translated to external entity labels, which can be used for Antrea NetworkPolicy configuration.

## Changes
New update handler is added in virtual machine rest handler to update the virtual machine in-memory cache. In case of restart, we use external entity object of the virtual machine to re-generate user tags.


pkg/apiserver/registry/inventory/virtualmachine/rest.go
```diff
@@ -83,6 +84,37 @@ func (r *REST) ShortNames() []string {
 	return []string{"vm"}
 }
 
+func (r *REST) Update(ctx context.Context, name string, objInfo rest.UpdatedObjectInfo, _ rest.ValidateObjectFunc,
```
Anandkumar26, line 87: nit: Add a comment saying only spec fields can be updated via the rest interface.

pkg/inventory/inventory.go
```diff
@@ -323,3 +323,9 @@ func (i *Inventory) compareVirtualMachineObjects(cached, discovered runtimev1alp
 	sortInterfacesFunc(dis.NetworkInterfaces)
 	return reflect.DeepEqual(cached.NetworkInterfaces, dis.NetworkInterfaces)
 }
+
+// UpdateVm updates virtual machine object in vm cache.
+func (i Inventory) UpdateVm(vm *runtimev1alpha1.VirtualMachine) error {
```
Anandkumar26, line 328: nit: func (i *Inventory)

Anandkumar26 (commented):
Minor comment, once addressed we can merge it.

pkg/apiserver/registry/inventory/virtualmachine/rest.go
reachjainrahul: its kind of expected.. Status is runtime, and Spec is desired.. So not required.

shenmo3 (approved):

PR #273, opened 2023-07-06
Avoid crashing nephe if CRD is not present

reachjainrahul:
## Description
CR controller tries to watch for CRs upon start. If the CRD definitions is not added by then, it use to restart the pod.

## Changes 
Added a routine to check for existence of CRDs before starting CR controller. It will loop for 1 hour with max backoff of 30 sec. After that nephe will be restarted.

### DIFF
--- a/build/charts/nephe/templates/controller/clusterrole.yaml
+++ b/build/charts/nephe/templates/controller/clusterrole.yaml
@@ -157,3 +157,9 @@ rules:
   verbs:
   - list
   - update
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - get
--- a/cmd/nephe-controller/main.go
+++ b/cmd/nephe-controller/main.go
@@ -39,6 +39,7 @@ import (
 	"antrea.io/nephe/pkg/controllers/virtualmachine"
 	"antrea.io/nephe/pkg/inventory"
 	"antrea.io/nephe/pkg/logging"
+	"antrea.io/nephe/pkg/util/k8s/crd"
 	// +kubebuilder:scaffold:imports
 )
 
@@ -94,6 +95,13 @@ func main() {
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
+		os.Exit(1)
+	}
+
+	// Wait for CRD to be present.
+	setupLog.Info("Checking for CRDs existence")
+	if err = crd.CheckCRDExistence(setupLog); err != nil {
+		setupLog.Info("failed to check for CRDs", "err", err)
 		os.Exit(1)
 	}
 
--- a/config/nephe.yml
+++ b/config/nephe.yml
@@ -450,6 +450,12 @@ rules:
   verbs:
   - list
   - update
+- apiGroups:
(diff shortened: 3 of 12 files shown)


pkg/util/crd/crd.go
```diff
@@ -0,0 +1,63 @@
...
+func CheckCRDExistence(log logr.Logger) error {
+	apiExtClient, err := apiextensionsclientset.NewForConfig(ctrl.GetConfigOrDie())
+	if err != nil {
+		return fmt.Errorf("failed to create API extensions client: %v", err)
+	}
+	b := backoff.NewExponentialBackOff()
+	// Configure the backoff parameters
+	b.InitialInterval = 1 * time.Second
+	b.MaxInterval = 30 * time.Second
+	b.MaxElapsedTime = 60 * time.Minute // Max retry
```
shenmo3, line 60: nit: 1 * time.Hour?

pkg/util/crd/crd.go
```diff
@@ -0,0 +1,63 @@
...
+
+	err = backoff.Retry(func() error {
+		crds := []string{
+			"cloudprovideraccounts.crd.cloud.antrea.io",
+			"cloudentityselectors.crd.cloud.antrea.io",
```
shenmo3, line 65: nit: define these as const or slice var at top at package level?

pkg/util/crd/crd.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+		for _, crd := range crds {
+			_, err := apiExtClient.ApiextensionsV1().CustomResourceDefinitions().Get(context.TODO(), crd, v1.GetOptions{})
+			if err != nil {
+				log.Info("failed to get CRD", "name", crd)
+				return fmt.Errorf("failed to get CRD: %v", err)
```
shenmo3, line 71: Do you also want to include crd name in error returned?

shenmo3 (commented):
Does it make more sense to put the crd.go under k8s directory?

pkg/util/crd/crd.go
reachjainrahul: err message will have the name

pkg/util/crd/crd.go
reachjainrahul: its actually in utility.. and relevant for this.. if it was dedicated package, then it make sense to put at the top. Hope it make sense

reachjainrahul:
> Does it make more sense to put the crd.go under k8s directory?

Sure.. we can put under k8s.. and move tags to new dir.. will make the change

shenmo3:
> > Does it make more sense to put the crd.go under k8s directory?
> 
> Sure.. we can put under k8s.. and move tags to new dir.. will make the change

I was thinking crd is also a part of k8s right, and having multiple directories each with just one file is a bit ugly? Just different styles I think, no need to change.

pkg/util/crd/crd.go
reachjainrahul: Done

shenmo3 (approved):
LGTM

reachjainrahul:
/nephe-test-e2e-kind,/nephe-test-e2e-agentless-upgrade

PR #275, opened 2023-07-11
Avoid retrying while syncing security groups

reachjainrahul:
## Description
Nephe runs periodic sync with cloud routine to sync cloud entity with internal cache. If any error occurs while syncing cloud entities, we want to avoid adding the entity in retry queue. This is to avoid unnecessary retries if sync with cloud fails, as it itself is another retry mechanism.

## Changes
With this PR, we maintain a new state on cloud entity to indicate if that entity is eligible for retry if any error occurs.


shenmo3 (dismissed):
LGTM

shenmo3 (approved):

PR #276, opened 2023-07-11
compute rule update delta against the cloud in plugins

shenmo3:
## Description
Currently, the plugin directly invokes cloud APIs with the update rule set passed from the network policy. However, this approach becomes problematic if the cloud rule indexer is out of sync with the actual cloud state, due to user actions or other failures. To address this issue, this PR introduces a solution that fetches and compares the current cloud rules with the update sets in the plugin before invoking cloud APIs. It computes the final sets of rules that need to be updated based on this comparison.

## Changes
1. Implement normalization for AWS returned rules and introduced logic for deduplication and finding duplicates. Deduplication is used to compare the add rule set and avoid adding rules that already exist in the cloud. Finding duplicates is used to compare the remove rule set and avoid removing rules that do not exist in the cloud.
2. Move rule conversion to a separate function instead of inside cloud API call function.
3. Implement deduplication checks for Azure to prevent adding rules that already exist in the cloud. The logic for finding duplicates already existed.
4. Move adding Azure default deny rule to separate func to avoid unnecessary comparisons in delta compute logic.

### DIFF
--- a/pkg/cloudprovider/plugins/aws/aws_converters.go
+++ b/pkg/cloudprovider/plugins/aws/aws_converters.go
@@ -15,6 +15,7 @@
 package aws
 
 import (
+	"fmt"
 	"net"
 	"strconv"
 	"strings"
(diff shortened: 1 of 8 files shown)


pkg/cloudprovider/plugins/aws/aws_converters.go
```diff
@@ -137,6 +138,63 @@ func convertFromSecurityGroupPair(cloudGroups []*ec2.UserIdGroupPair, managedSGs
 	return cloudResourceIDs, desc
 }
 
+func convertIngressToIpPermission(rules []*cloudresource.CloudRule, cloudSGNameToObj map[string]*ec2.SecurityGroup) (
```
Anandkumar26, line 141: nit: add a function description.

pkg/cloudprovider/plugins/aws/aws_converters.go
```diff
@@ -137,6 +138,63 @@ func convertFromSecurityGroupPair(cloudGroups []*ec2.UserIdGroupPair, managedSGs
...
+	[]*ec2.IpPermission, error) {
+	ipPermissions := make([]*ec2.IpPermission, 0)
+	for _, obj := range rules {
+		rule := obj.Rule.(*cloudresource.IngressRule)
+		if rule == nil {
+			continue
+		}
+		description, err := utils.GenerateCloudDescription(obj.NpNamespacedName)
+		if err != nil {
+			return nil, fmt.Errorf("unable to generate rule description, err: %v", err)
```
Anandkumar26, line 151: If any of the rules do not contain description, you fail the entire conversion?

shenmo3: This is generating description for Nephe rules, which shouldn't fail at all. In an error case, we do want to fail the entire conversion or operation so the cloud won't be in partial state.

pkg/cloudprovider/plugins/aws/aws_security.go
```diff
@@ -672,3 +625,76 @@ func getMemberNicCloudResourcesAttachedToOtherSGs(members []cloudresource.CloudR
...
+				ToPort:     ipPermission.ToPort,
+			}
+			normalizedList = append(normalizedList, normalized)
+		}
+		for _, group := range ipPermission.UserIdGroupPairs {
+			normalized := &ec2.IpPermission{
+				FromPort:         ipPermission.FromPort,
+				IpProtocol:       ipPermission.IpProtocol,
+				ToPort:           ipPermission.ToPort,
+				UserIdGroupPairs: []*ec2.UserIdGroupPair{{GroupId: group.GroupId, Description: group.Description}},
```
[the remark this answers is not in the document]
shenmo3: For groups we are constructing new `UserIdGroupPair` with only groupId and Description, ignoring other fields. For `IpRange` we are copying the object from previous slice into the new slice, which contains the description.

pkg/cloudprovider/plugins/aws/aws_security_impl.go
```diff
@@ -79,37 +79,61 @@ func (c *awsCloud) UpdateSecurityGroupRules(appliedToGroupIdentifier *cloudresou
...
+		return err
+	}
+	removeEgressRules, err := convertEgressToIpPermission(rmERule, cloudSGNameToCloudSGObj)
+	if err != nil {
+		return err
+	}
+
+	addIngressRules = dedupIpPermissions(addIngressRules, cloudSGObjToAddRules.IpPermissions)
+	addEgressRules = dedupIpPermissions(addEgressRules, cloudSGObjToAddRules.IpPermissionsEgress)
+	removeIngressRules = findDupIpPermissions(removeIngressRules, cloudSGObjToAddRules.IpPermissions)
```
reachjainrahul, line 104: Remove this for now..

shenmo3: Done

pkg/cloudprovider/plugins/azure/azure_security.go
```diff
@@ -420,24 +428,34 @@ func (computeCfg *computeServiceConfig) buildEffectivePeerNSGSecurityRulesToAppl
...
+				addAzureRules := addIngressRules
 				if *rule.Properties.Direction == armnetwork.SecurityRuleDirectionOutbound {
-					removeRules = rmEgressSecurityRules
+					removeAzureRules = rmEgressRules
+					addAzureRules = addEgressRules
 				}
-				normalizedRule := normalizeAzureSecurityRule(rule)
-				idx, found := findSecurityRule(normalizedRule, removeRules)
+				// skip the rule if found in remove list.
+				idx, found := findSecurityRule(normalizedRule, removeAzureRules)
```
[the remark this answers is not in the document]
shenmo3: Done

pkg/cloudprovider/plugins/aws/aws_converters.go
shenmo3: Done

reachjainrahul (approved):
/LGTM

PR #277, opened 2023-07-13
Add pagination to azure resource graph.

archanapholla:
Fix page size to 100 records.

### DIFF
--- a/pkg/cloudprovider/plugins/azure/azure_resource_graph_nwIntfTable.go
+++ b/pkg/cloudprovider/plugins/azure/azure_resource_graph_nwIntfTable.go
@@ -77,13 +77,9 @@ func getNetworkInterfaceTable(resourceGraphAPIClient azureResourceGraphWrapper, 
 	if err != nil {
 		return nil, 0, err
 	}
-	if data == nil {
-		return []*networkInterfaceTable{}, 0, nil
-	}
 
 	var networkInterfaces []*networkInterfaceTable
-	networkInterfaceRows := data.([]interface{})
-	for _, networkInterfaceRow := range networkInterfaceRows {
+	for _, networkInterfaceRow := range data {
 		var networkInterface networkInterfaceTable
 		err = mapstructure.Decode(networkInterfaceRow, &networkInterface)
 		if err != nil {
--- a/pkg/cloudprovider/plugins/azure/azure_resourcegraph.go
+++ b/pkg/cloudprovider/plugins/azure/azure_resourcegraph.go
@@ -16,8 +16,11 @@ package azure
 
 import (
 	"context"
+	"math"
 
 	resourcegraph "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resourcegraph/armresourcegraph"
+
+	"antrea.io/nephe/pkg/cloudprovider/plugins/internal"
 )
 
 const (
@@ -42,10 +45,16 @@ func (p *azureServiceSdkConfigProvider) resourceGraph() (azureResourceGraphWrapp
 }
 
 func invokeResourceGraphQuery(resourceGraphAPIClient azureResourceGraphWrapper, query *string,
-	subscriptions []*string) (interface{}, int64, error) {
+	subscriptions []*string) ([]interface{}, int64, error) {
+	var data []interface{}
+	var currentRecords int64
+	var totalRecords int64 = math.MaxInt64
+	var pageSize = int32(internal.MaxCloudResourceResponse)
+
 	resultFmt := resourcegraph.ResultFormatObjectArray
 	requestOptions := resourcegraph.QueryRequestOptions{
 		ResultFormat: &resultFmt,
+		Top:          &pageSize,
 	}
 
 	request := resourcegraph.QueryRequest{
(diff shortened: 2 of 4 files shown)


pkg/cloudprovider/plugins/azure/azure_resourcegraph.go
```diff
@@ -55,9 +62,18 @@ func invokeResourceGraphQuery(resourceGraphAPIClient azureResourceGraphWrapper,
...
 
-	results, queryErr := resourceGraphAPIClient.resources(context.Background(), request)
-	if queryErr == nil {
-		return results.Data, *results.TotalRecords, nil
+	// invoke resource graph API till all pages from response are fetched.
+	for currentRecords < totalRecords {
+		results, queryErr := resourceGraphAPIClient.resources(context.Background(), request)
+		if queryErr == nil {
+			data = append(data, results.Data)
+			currentRecords += *results.Count
```
reachjainrahul, line 70: Can't you use check if the result.Count is less than pageSize, break. This will avoid totalRecord variable?

pkg/cloudprovider/plugins/azure/azure_resourcegraph_vmTable.go
```diff
@@ -152,19 +152,24 @@ func getVirtualMachineTable(resourceGraphAPIClient azureResourceGraphWrapper, qu
...
 
 	var virtualMachines []*virtualMachineTable
-	virtualMachineRows := data.([]interface{})
-	for _, virtualMachineRow := range virtualMachineRows {
-		var virtualMachine virtualMachineTable
-		err = customDecode(virtualMachineRow.(map[string]interface{}), &virtualMachine)
-		if err != nil {
-			return nil, 0, fmt.Errorf("error decoding Azure resource graph query's response: %v", err)
+	for _, page := range data {
+		if page == nil {
```
reachjainrahul, line 161: What is the scenario when data is there, however page is nil? its not clear to me..

pkg/cloudprovider/plugins/azure/azure_resourcegraph.go
```diff
@@ -42,10 +43,16 @@ func (p *azureServiceSdkConfigProvider) resourceGraph() (azureResourceGraphWrapp
 }
 
 func invokeResourceGraphQuery(resourceGraphAPIClient azureResourceGraphWrapper, query *string,
-	subscriptions []*string) (interface{}, int64, error) {
+	subscriptions []*string) ([]interface{}, int64, error) {
+	var data = make([]interface{}, 0)
```
reachjainrahul, line 47: is this data slice of slice.. meaning [page1]....Vms [page2].... ???

pkg/cloudprovider/plugins/azure/azure_resourcegraph.go
```diff
@@ -42,10 +43,16 @@ func (p *azureServiceSdkConfigProvider) resourceGraph() (azureResourceGraphWrapp
...
+	var currentRecords int64
+	var totalRecords int64 = math.MaxInt64
+	var pageSize int32 = 100
```
reachjainrahul, line 50: Use this variable MaxCloudResourceResponse

pkg/cloudprovider/plugins/azure/azure_resource_graph_nwIntfTable.go
```diff
@@ -77,19 +77,24 @@ func getNetworkInterfaceTable(resourceGraphAPIClient azureResourceGraphWrapper,
...
 	}
 
 	var networkInterfaces []*networkInterfaceTable
-	networkInterfaceRows := data.([]interface{})
-	for _, networkInterfaceRow := range networkInterfaceRows {
-		var networkInterface networkInterfaceTable
-		err = mapstructure.Decode(networkInterfaceRow, &networkInterface)
-		if err != nil {
-			return nil, 0, err
+	for _, page := range data {
```
reachjainrahul, line 85: I wonder if you can avoid making changes in the caller, and let the logic of combining page data in invokeResourceGraphQuery api itself

pkg/cloudprovider/plugins/azure/azure_resourcegraph.go
archanapholla: I don't think we can use that, for result.Count exactly same as pageSize and totalRecord is also same as result.Count, we wouldn't want to enter the loop again. I think totalRecord is required.

pkg/cloudprovider/plugins/azure/azure_resource_graph_nwIntfTable.go
archanapholla: Logic in caller is specific to the type(virtulamachine/NetworkInterface), so decoding each page should stay in the caller itself.

pkg/cloudprovider/plugins/azure/azure_resourcegraph.go
archanapholla: Data returned by the API is expected to be an array(type is interface{} though as per query response), I am just appending this data in a slice and returning to the caller. Caller decodes each page.  Any comments here?

pkg/cloudprovider/plugins/azure/azure_resourcegraph_vmTable.go
archanapholla: As discussed, Updated the PR to have data field include the actual data instead of presenting in the format for individual page.

reachjainrahul (approved):
/LGTM

PR #278, opened 2023-07-21
Fix Secret Add event handling

Anandkumar26:
If a Secret Add event is received, and the reconciler has not yet received the CloudProviderAccount CR, the Secret handler should not process/replay the account.

Also, do not throw a reconciler error if a CloudEntitySelector CR is not present.


reachjainrahul (approved):
/LGTM

PR #279, opened 2023-07-21
Add explicit Antrea CRD dependency in Nephe

reachjainrahul:
## Description
Nephe controller has dependency on Antrea to be installed on the cluster first. If Antrea CRDs are not installed, Nephe controller will crash.

## Changes
Nephe controller will look for Antrea CRDs at boot, and if they are not installed, it will continue to loop and check for its existence in the cluster.


shenmo3 (approved):
LGTM
Do we also needs to check for ANP CRD?

reachjainrahul:
> LGTM Do we also needs to check for ANP CRD?

No, we dont consume ANP directly. just internal groups. however EE and EN is used for controller init, and the reason for crash

PR #280, opened 2023-07-21
Allow Nephe to talk to Antrea Controller directly

reachjainrahul:
## Description
Nephe controller talks to Antrea Controller API Server via K8s API Server. It has scale and performance issues, as the communication is via K8s Server.

## Changes
With this PR, we allow direct communication between Nephe controller and Antrea controller, if antreaKubeconfig is set in configMap.

### DIFF
--- a/pkg/controllers/networkpolicy/controller.go
+++ b/pkg/controllers/networkpolicy/controller.go
@@ -27,7 +27,9 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clientcmd"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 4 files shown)


pkg/controllers/networkpolicy/controller.go
```diff
@@ -603,7 +605,7 @@ func (r *NetworkPolicyReconciler) backgroupProcess() {
 }
 
 // SetupWithManager sets up NetworkPolicyReconciler with manager.
-func (r *NetworkPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
+func (r *NetworkPolicyReconciler) SetupWithManager(mgr ctrl.Manager, antreaKubeconfig string) error {
```
shenmo3, line 608: The func implements ImportReconciler, is it safe to remove this?

pkg/controllers/networkpolicy/controller.go
```diff
@@ -731,13 +733,31 @@ func (r *NetworkPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	if mgr == nil {
 		return nil
 	}
-	r.antreaClient = antreanetworkingclient.NewForConfigOrDie(mgr.GetConfig())
+
+	r.createAntreaClient(mgr, antreaKubeconfig)
```
shenmo3, line 737: imo its cleaner to let `createAntreaClient` return the client then set r.antreaClient here like before. This way the func also can be a normal func instead of member func of `NetworkPolicyReconciler`.

shenmo3 (approved):
LGTM with two minor comment

PR #282, opened 2023-07-24
Cleanup nephe manifests

reachjainrahul:
Removed service suffix from nephe-controller and nephe-controller-webhook service and also removed deleted unused virtualmachines file in crd as its moved to runtime object.

**Note:** This has upgrade implication. User would need to explicitly delete controller and webhook service before applying new yaml.


archanapholla (approved):
/LGTM

PR #286, opened 2023-07-27
Add a state in cloud account config

Anandkumar26:
The cloud account config contains the service configs and cloud api clients. Whenever cloud account config update fails, set the cloud account config state to invalid. 

Any consumer of cloud account config which makes the cloud API calls should check the cloud account config state and proceed only if it's in a valid state.

### DIFF
--- a/pkg/cloudprovider/plugins/internal/cloud_common.go
+++ b/pkg/cloudprovider/plugins/internal/cloud_common.go
@@ -39,6 +39,10 @@ var (
 	AccountCredentialsDefault       = "default"
 )
 
+const (
+	AccountConfigNotFound = "unable to find cloud account config"
+)
+
 type InstanceID string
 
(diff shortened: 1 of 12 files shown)


pkg/cloudprovider/plugins/internal/cloud_common.go
```diff
@@ -202,7 +205,9 @@ func (c *cloudCommon) DoInventoryPoll(accountNamespacedName *types.NamespacedNam
 	}
 	accCfg.LockMutex()
 	defer accCfg.UnlockMutex()
-
+	if !accCfg.GetAccountConfigState() {
```
reachjainrahul, line 208: Could you just update GetCloudAccountByName routine to return 'not found' or 'not valid' error.

Anandkumar26: Updated the function to return, account config, found and an error.

reachjainrahul (approved):
/LGTM

PR #289, opened 2023-08-03
implement multi region support for cpa

alexliu541:
## Description
This PR adds support for multi region configuration in CloudProviderAccount CRD.

## Changes
1. Change account config to have multiple service configs based on regions.
2. Modify webhook to validate all regions in the account config. For AWS it checks if region is supported or not, for Azure it only checks if its empty or not. Also added region validation in service config creation, where both AWS and Azure are checked if region is supported.
3. Adding handling to create AWS service config for each region and single Azure service config with multi-region filter.
4. Change account operations to support multiple service configs and run in parallel, such as inventory poll and cloud sync.
5. Add multi error in multi region operations, so success regions will be processed and error regions will be reported separately.
6. Change security group operations to retrieve regions using vpc/vnet IDs instead of from cloud account.
7. Change Azure cloud sync process to use ASG ID internally when syncing to avoid ASGs with the same name in different region being grouped into one.
8. Update unit-tests to work with multi region.

### DIFF
--- a/pkg/accountmanager/poller.go
+++ b/pkg/accountmanager/poller.go
@@ -148,7 +148,7 @@ func (p *accountPoller) doAccountPolling() {
 	}
 
 	// TODO: Remove this when event based push is implemented in the plugin.
-	cloudInventory, err := p.cloudInterface.GetCloudInventory(p.accountNamespacedName)
+	cloudInventory, err := p.cloudInterface.GetAccountCloudInventory(p.accountNamespacedName)
 	if err != nil {
 		// Chances are while polling was happening, account is removed.
 		return
--- a/pkg/apiserver/webhook/cloudprovideraccount_webhook.go
+++ b/pkg/apiserver/webhook/cloudprovideraccount_webhook.go
@@ -39,7 +39,7 @@ import (
 
 const MinPollInterval = 30
 
-var (
+const (
 	errorMsgSecretNotConfigured    = "unable to get secret"
 	errorMsgMinPollInterval        = "pollIntervalInSeconds should be >= 30. If not specified, defaults to 60"
 	errorMsgMissingAwsCredential   = "must specify either credentials or role arn, both cannot be empty"
@@ -248,20 +248,21 @@ func (v *CPAValidator) validateAWSAccount(account *crdv1alpha1.CloudProviderAcco
 		return fmt.Errorf(errorMsgMissingAwsCredential)
 	}
 
-	if len(awsConfig.Region) == 0 || len(strings.TrimSpace(awsConfig.Region[0])) == 0 {
+	if len(awsConfig.Region) == 0 {
 		return fmt.Errorf(errorMsgMissingRegion)
 	}
 
 	// NOTE: currently only AWS standard partition regions supported (aws-cn, aws-us-gov etc are not
 	// supported). As we add support for other partitions, validation needs to be updated.
-	regions := endpoints.AwsPartition().Regions()
-	_, found := regions[awsConfig.Region[0]]
-	if !found {
-		var supportedRegions []string
-		for key := range regions {
-			supportedRegions = append(supportedRegions, key)
+	awsRegionMap := endpoints.AwsPartition().Regions()
+	for _, region := range awsConfig.Region {
+		if _, found := awsRegionMap[strings.ToLower(region)]; !found {
+			var supportedRegions []string
+			for key := range awsRegionMap {
+				supportedRegions = append(supportedRegions, key)
+			}
+			return fmt.Errorf("%v %s [%v]", region, errorMsgInvalidRegion, supportedRegions)
 		}
-		return fmt.Errorf("%v %s [%v]", awsConfig.Region, errorMsgInvalidRegion, supportedRegions)
 	}
 
 	return nil
@@ -310,10 +311,14 @@ func (v *CPAValidator) validateAzureAccount(account *crdv1alpha1.CloudProviderAc
 	if len(strings.TrimSpace(azureCredential.ClientKey)) == 0 && len(strings.TrimSpace(azureCredential.SessionToken)) == 0 {
 		return fmt.Errorf(errorMsgMissingAzureCredential)
 	}
-
 	// validate region
-	if len(azureConfig.Region) == 0 || len(strings.TrimSpace(azureConfig.Region[0])) == 0 {
+	if len(azureConfig.Region) == 0 {
 		return fmt.Errorf(errorMsgMissingRegion)
+	}
+	for _, region := range azureConfig.Region {
+		if len(strings.TrimSpace(region)) == 0 {
+			return fmt.Errorf(errorMsgMissingRegion)
+		}
 	}
 
 	return nil
--- a/pkg/cloudprovider/plugins/aws/aws_account_impl.go
+++ b/pkg/cloudprovider/plugins/aws/aws_account_impl.go
@@ -15,6 +15,8 @@
 package aws
 
 import (
+	"strings"
+
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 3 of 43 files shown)


pkg/cloudprovider/plugins/aws/aws_account_mgmt.go
```diff
@@ -82,14 +84,16 @@ func compareAccountCredentials(accountName string, existing interface{}, new int
...
-	if strings.Compare(existingConfig.region, newConfig.region) != 0 {
-		credsChanged = true
-		awsPluginLogger().Info("Account region updated", "account", accountName)
-	}
 	if strings.Compare(existingConfig.endpoint, newConfig.endpoint) != 0 {
 		credsChanged = true
 		awsPluginLogger().Info("Endpoint url updated", "account", accountName)
 	}
+	sort.Strings(existingConfig.regions)
+	sort.Strings(newConfig.regions)
```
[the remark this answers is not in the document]
shenmo3: I actually added to lower in aws_account_impl.go and azure_account_impl.go. This converts regions to lower before any processing so we don't need to worry about to lower in actual logics. However regions from the cloud, either in VM, VNET or region list still needs to be converted to lower.

pkg/apiserver/webhook/cloudprovideraccount_webhook.go
```diff
@@ -247,20 +247,22 @@ func (v *CPAValidator) validateAWSAccount(account *crdv1alpha1.CloudProviderAcco
...
 	// NOTE: currently only AWS standard partition regions supported (aws-cn, aws-us-gov etc are not
 	// supported). As we add support for other partitions, validation needs to be updated.
-	regions := endpoints.AwsPartition().Regions()
-	_, found := regions[awsConfig.Region[0]]
-	if !found {
-		var supportedRegions []string
-		for key := range regions {
-			supportedRegions = append(supportedRegions, key)
+	awsRegionMap := endpoints.AwsPartition().Regions()
+	awsRegionMap["all"] = endpoints.Region{}
```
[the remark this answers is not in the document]
shenmo3: Done

pkg/cloudprovider/plugins/aws/aws_account_impl.go
```diff
@@ -23,6 +25,9 @@ import (
 
 // AddProviderAccount adds and initializes given account of a cloud provider.
 func (c *awsCloud) AddProviderAccount(client client.Client, account *crdv1alpha1.CloudProviderAccount) error {
+	for idx := range account.Spec.AWSConfig.Region {
```
Anandkumar26, line 28: Modifying the retrieved CR here, seems inappropriate.
Going forward, we may consume the region field before calling this function.

pkg/cloudprovider/plugins/aws/aws_api_wrappers.go
```diff
@@ -144,3 +150,8 @@ func (ec2Wrapper *awsEC2WrapperImpl) describeVpcPeeringConnectionsWrapper(input
 	*ec2.DescribeVpcPeeringConnectionsOutput, error) {
 	return ec2Wrapper.ec2.DescribeVpcPeeringConnections(input)
 }
+
+// getRegion returns the region of the ec2 client.
+func (ec2Wrapper *awsEC2WrapperImpl) getRegion() string {
+	return *ec2Wrapper.ec2.Config.Region
```
Anandkumar26, line 156: Since you are dereferencing the pointer, check if there id any scenario when region could be empty.

pkg/cloudprovider/plugins/aws/aws_cloudcommonhelper_impl.go
```diff
@@ -22,10 +22,14 @@ func (h *awsCloudCommonHelperImpl) GetCloudServicesCreateFunc() internal.CloudSe
...
+	return updateAwsServiceConfigs
 }
 
-func (h *awsCloudCommonHelperImpl) GetCloudCredentialsComparatorFunc() internal.CloudCredentialComparatorFunc {
-	return compareAccountCredentials
+func (h *awsCloudCommonHelperImpl) SetAccountConfigFunc() internal.CloudConfigValidatorFunc {
+	return setAccountConfig
+}
+
+func (h *awsCloudCommonHelperImpl) GetCloudConfigComparatorFunc() internal.CloudConfigComparatorFunc {
```
Anandkumar26, line 33: nit: 
```suggestion
func (h *awsCloudCommonHelperImpl) GetAccountConfigComparatorFunc() internal.CloudConfigComparatorFunc {
```

pkg/cloudprovider/plugins/aws/aws_ec2.go
```diff
@@ -51,27 +51,36 @@ type ec2ResourcesCacheSnapshot struct {
...
+// and multierr for any errors.
+func newEC2ServiceConfig(accountNamespacedName types.NamespacedName, services map[string]awsServiceClientCreateInterface) (
+	map[string]internal.CloudServiceInterface, error) {
+	var retErr error
+	configs := make(map[string]internal.CloudServiceInterface)
+	for region, service := range services {
+		// create ec2 sdk api client.
+		apiClient, err := service.compute()
+		if err != nil {
+			err = fmt.Errorf("error creating ec2 sdk api client for account: %s, region: %s, err: %v", accountNamespacedName.String(),
```
Anandkumar26, line 64: Can you try/simulae a scenario where this function returns a multi error (i.e. multiple errors) and check how it would look on the CPA status field?

pkg/cloudprovider/plugins/aws/aws_services.go
```diff
@@ -107,20 +112,87 @@ func (h *awsServicesHelperImpl) newServiceSdkConfigProvider(accConfig *awsAccoun
...
+// newAwsServiceConfigs creates service configs for each region for an account and returns a region to service config map for successful
+// creations and multierr for any errors.
+func newAwsServiceConfigs(accountNamespacedName *types.NamespacedName, config interface{}, awsSpecificHelper interface{}) (
+	map[string]internal.CloudServiceInterface, error) {
 	awsServicesHelper := awsSpecificHelper.(awsServicesHelper)
-	awsAccountCredentials := accCredentials.(*awsAccountConfig)
+	awsConfig := config.(*awsAccountConfig)
 
-	awsServiceClientCreator, err := awsServicesHelper.newServiceSdkConfigProvider(awsAccountCredentials)
+	// check regions valid and process regions.
```
Anandkumar26, line 122:
```suggestion
	// check valid regions.
```

pkg/accountmanager/manager.go
```diff
@@ -223,7 +223,9 @@ func (a *AccountManager) RemoveResourceFiltersFromAccount(accNamespacedName *typ
 		return fmt.Errorf(fmt.Sprintf("failed to delete selector %v, account %v: %v",
 			selectorNamespacedName, accNamespacedName, errorMsgAccountPollerNotFound))
 	}
+	accPoller.mutex.Lock()
```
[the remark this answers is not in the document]
shenmo3: As we discussed offline, there is a race condition that needs this lock to be avoided. Will move the lock to the top of the func to make sure we are in lock when modifying anything.

pkg/accountmanager/manager.go
```diff
@@ -285,7 +287,9 @@ func (a *AccountManager) removeAccountPoller(namespacedName *types.NamespacedNam
 	delete(a.accPollers, *namespacedName)
 
 	accPoller.stopPoller()
+	accPoller.mutex.Lock()
```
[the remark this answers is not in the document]
shenmo3: This part actually doesn't need a lock since the poller is already stopped, will remove it.

pkg/accountmanager/poller.go
```diff
@@ -145,7 +145,7 @@ func (p *accountPoller) doAccountPolling() {
 		p.updateAccountStatus(p.cloudInterface)
 	}()
 
-	cloudInventory, err := p.cloudInterface.GetCloudInventory(p.accountNamespacedName)
+	cloudInventory, err := p.cloudInterface.GetAccountCloudInventory(p.accountNamespacedName)
```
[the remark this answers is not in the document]
shenmo3: I changed it because service config also has func `GetCloudInventory`, but thats just for a single service config. Since account config can now have multiple service configs and this func combines all service configs' inventory, I want to change this func name to differentiate the two.

Don't quite like ByAccount also, since we are already got a specific account config, how about `GetCloudInventoryForAccount`?

pkg/cloudprovider/plugins/aws/aws_api_wrappers.go
shenmo3: I'll add a check, but for AWS clients, the region is a required field.

reachjainrahul (approved):
/LGTM

PR #295, opened 2023-08-16
Improve account failure state handling and logging

reachjainrahul:
Store account failure state in account config struct inside plugin so that CPA/CES status can be updated with correct info. Also changed logging of some error logs from info to error.

### DIFF
--- a/pkg/cloudprovider/plugins/internal/accounts_common.go
+++ b/pkg/cloudprovider/plugins/internal/accounts_common.go
@@ -78,40 +78,45 @@ func (c *cloudCommon) newCloudAccountConfig(client client.Client, namespacedName
 		return nil, err
 	}
 
-	cloudConvertedCredential, err := credentialsValidatorFunc(client, credentials)
-	if err != nil {
-		return nil, err
-	}
-
-	serviceConfig, err := cloudServicesCreateFunc(namespacedName, cloudConvertedCredential, c.cloudSpecificHelper)
-	if err != nil {
-		return nil, err
-	}
-
-	status := &crdv1alpha1.CloudProviderAccountStatus{}
-	return &cloudAccountConfig{
+	accConfig := &cloudAccountConfig{
 		logger:         loggerFunc,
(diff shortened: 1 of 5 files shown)


reachjainrahul:
nephe-test-e2e-agentless-upgrade

pkg/cloudprovider/plugins/internal/accounts_common.go
```diff
@@ -137,9 +142,13 @@ func (c *cloudCommon) updateCloudAccountConfig(client client.Client, credentials
 		return err
 	}
 
-	if err := currentConfig.serviceConfig.UpdateServiceConfig(serviceConfig); err != nil {
-		cloudAccountConfigState = false
-		return err
+	if currentConfig.serviceConfig == nil {
+		currentConfig.serviceConfig = serviceConfig
+	} else {
```
shenmo3, line 147: Consider move the update outside of the else block and just return nil after L146 in the if block.

pkg/cloudprovider/plugins/internal/accounts_common.go
```diff
@@ -78,40 +78,45 @@ func (c *cloudCommon) newCloudAccountConfig(client client.Client, namespacedName
...
-		// Update the cloud account config state for each update.
-		currentConfig.state = cloudAccountConfigState
-	}()
 
 	credentialsValidatorFunc, cloudServicesCreateFunc, credentialsComparatorFunc, err := c.getFunctionPointers()
 	if err != nil {
 		cloudAccountConfigState = false
 		return err
 	}
+	defer func() {
```
shenmo3, line 113: we don't need to update state if error happen on getting func pointers?

pkg/cloudprovider/plugins/internal/accounts_common.go
reachjainrahul: We want to continue if no error occurs. For example, if someone adds some code at line 153, that won't be executed.

shenmo3 (approved):

PR #308, opened 2023-09-06
Fixes to entity selector integration test.

archanapholla:
- Reduce wait time after CES CR config is applied.
- Fetch existing CES before updating the CES from the test in order to get the latest version of CES CR.


reachjainrahul (approved):
/LGTM

PR #309, opened 2023-09-13
Fix issue with ANP having no from field

reachjainrahul:
## Description
Handled 2 bugfix

## Changes
- In case of no 'from' field in ANP, the normalization of cloudrule was not matching anp rule. Reason, the protocol field wasn't set to "*". It was set to "", due to which cloudrule was not removed.
- Secondly, when addressgroup id and applied id are same, and if addressgroup is getting removed, it was clearing rules from cloudruleindexer for appliedTo group. Added AT check if prevent that from happening

### DIFF
--- a/pkg/cloudprovider/plugins/azure/azure_nsg_rules.go
+++ b/pkg/cloudprovider/plugins/azure/azure_nsg_rules.go
@@ -427,7 +427,10 @@ func findSecurityRule(ruleList []*armnetwork.SecurityRule, rule *armnetwork.Secu
 func normalizeAzureSecurityRule(rule *armnetwork.SecurityRule) *armnetwork.SecurityRule {
 	property := *rule.Properties
 	normalizedProtocolNum := azureProtoNameToNumMap[strings.ToLower(string(*rule.Properties.Protocol))]
-	normalizedProtocol := protoNumAzureNameMap[normalizedProtocolNum]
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/plugins/azure/azure_nsg_rules.go
```diff
@@ -427,7 +427,10 @@ func findSecurityRule(ruleList []*armnetwork.SecurityRule, rule *armnetwork.Secu
 func normalizeAzureSecurityRule(rule *armnetwork.SecurityRule) *armnetwork.SecurityRule {
 	property := *rule.Properties
 	normalizedProtocolNum := azureProtoNameToNumMap[strings.ToLower(string(*rule.Properties.Protocol))]
-	normalizedProtocol := protoNumAzureNameMap[normalizedProtocolNum]
+	normalizedProtocol, ok := protoNumAzureNameMap[normalizedProtocolNum]
+	if !ok {
+		normalizedProtocol = "*"
```
shenmo3, line 432: we should have specific mapping of "*" to "*" in conversion. For not found cases we should give an info log as it means a user rule is in nephe priority range with nephe description. (This func's caller already did priority and description check.) We can also try to remove this user rule by returning nil here and let the caller check nil and remove the rule

shenmo3 (approved):

PR #316, opened 2023-11-08
Expose ANP priority in test templates for Agented VMs

Anandkumar26:
Use ubuntu Vms for Azure agented tests
On Azure RHEL VM, when ExternalNode is added and deleted
repeatedly, antrea-agent goes into a weired state where
ExternalNode add event is ignored, as agent things there
is no change in the ExternalNode interface.

Tracking issue on antrea:

[#5192](https://github.com/antrea-io/antrea/issues/5192)
[#5111](https://github.com/antrea-io/antrea/issues/5111)


reachjainrahul (approved):
/LGTM

